PR #134, opened 2015-01-22
Added Yast::RSpec::SCR extension

ancorgs:


README.md
```diff
@@ -109,8 +109,28 @@ content = ButtonBox(
...
+The YaST team encourages to use RSpec for testing YaST code in Ruby. To help in
+that task, this gem includes some RSpec extensions. In order to use them all,
+the following line must be added to the tests.
+
+```ruby
+require 'yast/rspec'
+```
+
+For example, the following code makes use of the #path helper provided by
+{Yast::RSpec::SCR} (the only extension currently provided).
```
jreidinger: I think when we add another helps we forgot to remove this note, so I think we should add it there as it do not provide anything usefull. Maybe better to point to rubydoc page for documentaiton for all rspec helpers.

src/ruby/yast/rspec/scr.rb
```diff
@@ -0,0 +1,84 @@
...
+      end
+
+      # Encapsulates subsequent SCR calls into a chroot.
+      #
+      # Raises an exception if something goes wrong.
+      #
+      # @param [#to_s] directory to use as '/' for SCR calls
+      def change_scr_root(directory)
+        # On first call, store the default handler in the stack
+        @scr_handles ||= [Yast::WFM.SCRGetDefault]
```
jreidinger: in general I think here would be really nice if we can detect that `@scr_handles.size == 1` otherwise it means that some test forgot to close its SCR instance and cause problems for other tests. Another option is to automagically hook to after closing, that ensure it.

ancorgs: The current implementation is actually a stack, so it would support any level of nested chrooting. We can asume that we never want such thing and drop the ```@scr_handles``` array in favor of two scalars ```@scr_handle``` and ```@previous_scr_handle``` so we raise an exception if ```change_scr_root``` is called twice without calling ```reset_scr_root``` in the meantime.

I don't get the hook part of your comment, can you elaborate?

jreidinger: well, I think in rspec you can define own after, which search if `@scr_handles` size is bigger then one and automatic call reset_scr to avoid mistakes

jreidinger: and having it as stack looks like overengineering for me. What is use case? I think it do not make sense to have in stack more then one scr handle there

ancorgs: @jreidinger I was thinking one step further, like making this possible (I'm not 100% sure it can be implemented)

```ruby
context "in a xx system", scr_root: "xx" do
  it {true}
end
```

But the problem that I see is that the name of the chroot is usually not that short and straightforward as "xx" but most likely something like ```File.join(File.dirname(__FILE__), "data", "xx")```. Too large for an scenario condition.

jreidinger: I see different problem. Context is written to output, but it is test details that it run in xx fake system and do not say anything about described method. So I am not much convinced. One possibility is to use helper method with block, so no usual context, but it adds nesting, so I am not much convinced

ancorgs: @jreidinger On the other hand, I don't like the idea of automagically calling reset in the after hook because I can imagine situations in which it does the wrong thing. So I think I will.

1) Remove the stack behavior, supporting just one level.
2) Add the check to prevent two consecutive calls to ```change_scr_root``` without a reset.
3) Allow ```change_scr_root``` to be called with a block. So it's up to the user, if we want to explicitly call reset (most likely with a ```change_scr_root``` on before and a ```reset_scr_root``` on after) or if he want to use it in block form (much like File.open in which you have both options).

ancorgs:
Updated with improvements inspired by @jreidinger comments. Now ```change_src_root``` supports being called with a block and prevents opening a chroot if another one is open. Tests included.

ancorgs:
Re-reading I'm not sure if defining the ```path``` method in a extension called ```SCR``` makes 100% sense. Should I extract ```path``` to another extension? Should I rename the extension to something more generic like ```AgentHelpers```?

src/ruby/yast/rspec/scr.rb
```diff
@@ -0,0 +1,96 @@
...
+      # Nesting of chroots is forbidden and the method will raise an exception
+      # if is called without closing a previous chroot.
+      #
+      # @param directory [#to_s] directory to use as '/' for SCR calls
+      #
+      # @example Usage with a block
+      #   change_scr_root("/home/chroot1") do
+      #     # This reads the content of /home/chroot1/
+      #     Yast::SCR.Read(path(".target.dir"), "/")
+      #   end
```
jreidinger: Nice, I think we should also add example with rspec like:
```
around(:all) { |example| change_scr_root("/home/chroot1", &example) }
```
this way it is safe and easy to use :)

jreidinger:
@ancorgs regarding `path` I agree that own namespace would be nice and maybe better name is `Shortcuts` and include term and path and maybe in future other common stuff, that is repeated and is too long for test that it deserve shorting.

jreidinger:
LGTM

PR #136, opened 2015-01-28
Improve shortcuts

jreidinger:
@ancorgs BTW you do not announce it yet on yast-devel mailing list, it would be nice to do it.


ancorgs:
LGTM

PR #137, opened 2015-02-02
Run tests at Travis

lslezak:
- for some reason `LC_ALL` needs to be set when testing locale dependent code
- disable some `crypt()` tests (do not work in Ubutu, unfortunately this is deep in glibc and cannot be easily fixed)


jreidinger:
LGTM, thanks for fixing

PR #138, opened 2015-02-24
Fix ruby 22

jreidinger:


ancorgs:
Not my area of expertise, but it LGTM

PR #139, opened 2015-03-05
Fix building with yast2-core-3.1.16: use C++11 like core does

mvidner:
https://bugzilla.novell.com/show_bug.cgi?id=914255


jreidinger:
LGTM

PR #140, opened 2015-03-19
First experiment to implement rails-like core extensions

ancorgs:
While discussing https://github.com/yast/yast-support/pull/10 (a quite trivial bugfix) we started to discuss that would be the best place to place a method to clean ANSI characters from a string. We ended up thinking in ActiveSupport's CoreExtensions (nice example of the sentence "one thing leads to another").

This is some kind of reimplementation of the CoreExtensions concept. We can, of course decide this is not the right place or the right way. Even if we decide to keep the idea and to implement it here in ruby-binding, we should probably change the implementation to use Ruby refinements.


src/ruby/yast/core_ext/string/ansi.rb
```diff
@@ -0,0 +1,18 @@
+class String
+  # Removes color and cursor escape codes (modifying the string)
+  # http://en.wikipedia.org/wiki/ANSI_escape_code
+  def remove_ansi_sequences
+    remove_ansi_color_sequences
+    remove_ansi_cursor_sequences
+  end
+
+  # Removes ANSI color escape codes (modifying the string)
+  def remove_ansi_color_sequences
```
jreidinger: do we need it public?

src/ruby/yast/core_ext.rb
```diff
@@ -0,0 +1 @@
+require_relative "core_ext/string"
```
jreidinger: well, I know this is active support way, another option is module way, when you create module and if you need then use
```
a = "abc"
a.extend AnsiCleaning
a.remove_ansi_sequences
```
Advantage of the second way is that string class is not so fat. Also it is less magic. Disadvantage is that you need another call.

src/ruby/yast/core_ext/string/ansi.rb
ancorgs: It does not hurt, IMO. I guess you always want to remove all ANSI sequences (color and cursor), but maybe sometimes you are only interested in one type. But I'm not strongly against offering only ```remove_ansi_sequences```

src/ruby/yast/core_ext.rb
ancorgs: Hmmm, in the particular use case that originated this PR (just cleaning a single string) I like the suggested approach of using ```extend```. But if we think about other use cases for the core extensions idea, it's probably too cumbersome (i.e., not that convenient anymore). To prevent String to become fat and eventually dangerous I'd simply use refinements. I don't find it too much magical, you have to explicitly require the extension anyways.

jreidinger: problem is if you require extension in any file, then it is in all files everywhere (if you mean clean string extension not refinement). Any of problem with this extending class is that it can also cause collision. Consider that you use in an yast module another core extension, that have same method, just with slightly different implementation and then it is only on execution order which one is used and can cause very nasty effects.
BTW another approach, I like maybe even more and is more object oriented is to create
```
AnsiString < ::String
  def remove_ansi_sequences
  ...
  ...
end
```
so it is clear, that you have instance of string with ansi characters and it have own methods.
About refinements, I think it is quite good solution, just it quite new, so I don't yet have much experience about it, so do not know yet weak sides.

src/ruby/yast/core_ext/string/ansi.rb
jreidinger: `but maybe sometimes` => over-engineering
I propose to have it private and if in future such demand appear, then make it public with proper version dump. Otherwise we may find in future, that we need something slightly different then just `remove_ansi_color_sequences`

src/ruby/yast/core_ext.rb
ancorgs: I don't think we have to worry about different core extensions creating the same method. The set of extensions will be always kept under control, I don't expect we are going to get mad creating a lot of extensions.

I don't fully like the inheritance approach. It makes impossible to combine two different "core extensions".

I'll try to adapt this PR to use ruby refinements. After considering the options it still look like the way to go for me.

jreidinger: I agree that ruby refinements is probably the best way to go

src/ruby/yast/core_ext/ansi_string.rb
```diff
@@ -0,0 +1,26 @@
+module Yast
+  module CoreExt
+    module AnsiString
```
jreidinger: looks good, just please add documentation for module and also examples of usage as it is first time we use refinements in yast.

jreidinger:
LGTM

PR #141, opened 2015-04-01
Rspec matchers

ancorgs:


jreidinger:
in general it looks good, just how will look error message if failed?

ancorgs:
@jreidinger Fixed, now error messages look like this:
```
Yast::RSpec::Matchers#path_matching does not match
Failure/Error: Yast::SCR.Read(path(".etc"))
  Yast::SCR received :Read with unexpected arguments
    expected: (path_matching(/target.*ir$/))
         got: (#<Yast::Path:0x00000001c07cd0 @components=["etc"]>)
```

jreidinger:
yep, looks better

jreidinger:
just tests do not pass in travis

src/ruby/yast/rspec/matchers.rb
```diff
@@ -0,0 +1,40 @@
...
+      end
+
+      # @private
+      class PathMatchingMatcher
+        def initialize(expected)
+          @expected = Regexp.new(expected)
+        end
+
+        # RSpec 3 uses === while RSpec 2 uses ==
+        # Thus, implementing just == should work with both
```
jreidinger: well, maybe add FIXME as it actually is not `==` and `===` is really proper method regarding behavior of class. but just nitpick

jreidinger:
LGTM

PR #142, opened 2015-05-18
 Init the UI without y2base (bsc#922023)

mvidner:
https://bugzilla.suse.com/show_bug.cgi?id=922023


jreidinger:
looks good code wise,. but I missing documentation, ideally with examples, as I have no idea what to do with setUI and what it affect.

mvidner:
I have added doc comments, and here is a sample script: https://gist.github.com/mvidner/e99aec71c0e34c2a70ff

Note that there is no detection like in /sbin/yast. It is a low level API, I think that it is enough for the testing use case.

For further steps, we may want to move the UI detection logic from Bash to Ruby.

BTW I wasn’t able to render the C method documentation with Yard. Has anyone succeeded? It worked with RDoc.

jreidinger:
@mvidner I have no success with Yard.
Just please move example from gist to code documentation, noone find it in gist few years later, but it is quite good short example how to use it.

Thanks

src/binary/Yast.cc
```diff
@@ -384,6 +385,97 @@ static VALUE code_call( int argc, VALUE *argv, VALUE self )
...
+ *    if Yast.ui_component == ""
+ *      Yast.ui_component = ARGV[0] || "ncurses"
+ *    end
+ *
+ *    Builtins.y2milestone("UI component: %1", Yast.ui_component)
+ *    Yast.import "UI"
+ *
+ *    UI.OpenDialog(PushButton("This is a button"))
+ *    UI.UserInput
+ *    UI.CloseDialog
```
mvidner: Example added ^

jreidinger:
LGTM

PR #143, opened 2015-05-25
Local scr

jreidinger:


ancorgs:
Not sure if "local" express exactly the idea. Since I don't have a better suggestion, I'd probably change the direction of the question and would implement a ```#chroot?``` method.

jreidinger:
so something like `#scr_chroot?` as I want emphasis that it is part of SCR, as WFM currently handling three different functonality and I cannot split it due to backward compatibility

ancorgs:
Sounds good, yes.

jreidinger:
@ancorgs updated

ancorgs:
LGTM.

BTW, I'm not allowed to access the mentioned FATE, so I expect not many people being able.

PR #144, opened 2015-05-26
Fixed a strdup/delete mismatch (boo#932306).

mvidner:
The manual says it is actually harmless.

Mismatched free() / delete / delete []
http://valgrind.org/docs/manual/mc-manual.html#mc-manual.rudefn

https://bugzilla.suse.com/show_bug.cgi?id=932306


jreidinger:
thanks, LGTM

PR #145, opened 2015-06-03
Fixed "Comparable#== will no more rescue exceptions of #<=>" (bsc#933470)

mvidner:
https://bugzilla.suse.com/show_bug.cgi?id=933470

Term and Path would raise exceptions when <=> was used to compare them
with something that was not a Term or Path.

In particular it happened during RSpec argument matching.

In older Ruby core, Comparable#== would silently mask such exceptions,
now it started producing warnings.
https://bugs.ruby-lang.org/issues/7688

### DIFF
--- a/tests/ruby/term_spec.rb
+++ b/tests/ruby/term_spec.rb
@@ -1,10 +1,6 @@
 #!/usr/bin/env rspec
 # encoding: utf-8
 
-# FIXME: this file was autoconverted from test/unit syntax without
-# adjusting it to good RSpec style (http://betterspecs.org/).
-# Please improve it whenever adding examples.
-
 require_relative "test_helper"
 
 require "yast/term"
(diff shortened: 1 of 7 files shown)


tests/ruby/term_spec.rb
```diff
@@ -75,6 +71,10 @@ def term(*params)
     it "if value is equal, then use params to comparison" do
       expect(term(:HBox, "test")).to be > term(:HBox)
     end
+
+    it "if non-term, then uncomparable" do
```
jreidinger: maybe be specific like like `then return nil`?

tests/ruby/term_spec.rb
```diff
@@ -75,6 +71,10 @@ def term(*params)
...
+      expect(term(:HBox, "test").<=> 42).to eq nil
```
jreidinger: dot is not needed. try 
```
jreidinger@pepa:~/yast/bootloader/package> irb
irb(main):001:0> 1 <=> 3
=> -1
```

jreidinger:
Thanks for fixing, just two nitpickings.

PR #149, opened 2015-06-25
Builtins::strftime

ancorgs:
Apart from the FIXME comment it should be ready to merge if the C++ gurus find the call to ```locale::global``` acceptable.


src/binary/Builtin.cc
```diff
@@ -632,6 +632,39 @@ extern "C" {
     return INT2FIX(strcoll(RSTRING_PTR(str1), RSTRING_PTR(str2)));
   }
 
+  static struct tm *
+  hash_to_tm(VALUE hash)
+  {
+    struct tm res;
```
mvidner: `static struct tm res;`

src/binary/Builtin.cc
```diff
@@ -632,6 +632,39 @@ extern "C" {
...
+
+    res.tm_sec    = rb_hash_aref(hash, ID2SYM(rb_intern("tm_sec")));
```
mvidner: You get a VALUE and treat it as an int. I think it should be: 
```
res.tm_sec = NUM2INT(rb_hash_aref(...));
```

http://ruby-doc.com/docs/ProgrammingRuby/html/ext_ruby.html#UB

ancorgs: Indeed. And that explain why I was getting "dic 51 37:73:103" instead of "jun 25 18:41:19" :-)

src/binary/Builtin.cc
```diff
@@ -632,6 +632,45 @@ extern "C" {
...
+  static int
+  tm_hash_value(VALUE hash, const char *key)
+  {
+    return NUM2INT(rb_hash_aref(hash, ID2SYM(rb_intern(key))));
+  }
+  
+  static struct tm *
+  hash_to_tm(VALUE hash)
+  {
+    static struct tm res;
```
jreidinger: well, mvidner suggest to use static, which is ok, if you want to survive your pointer, but on other hand, this statis is not thread safe, so if we start using threads in yast, then we have problem if two threads call strftime method. Maybe we should consider a bit slower, but safer passing by value. or pass tm struct as reference parameter that was filled

src/binary/Builtin.cc
```diff
@@ -632,6 +632,45 @@ extern "C" {
...
+    char res[100];
+    struct tm *timeinfo;
+
+    Check_Type(format, T_STRING);
+    timeinfo = hash_to_tm(time);
+
+    if (strftime(res, sizeof(res), RSTRING_PTR(format), timeinfo)) {
+      return yrb_utf8_str_new(string(res));
+    } else {
+      return Qnil; // TODO: learn how to raise an exception
```
jreidinger: exception is raised e.g. this way: https://github.com/yast/yast-ruby-bindings/blob/7f6545d5185aa3d3d49d907b5df08c3859d9c83b/src/binary/Yast.cc#L120 or https://github.com/yast/yast-ruby-bindings/blob/7f6545d5185aa3d3d49d907b5df08c3859d9c83b/src/binary/Yast.cc#L88

src/ruby/yast/yast.rb
```diff
@@ -197,6 +197,37 @@ def self.#{sname}= (value)
     base.const_set(modules.last, m) unless base.constants.include?(modules.last.to_sym)
   end
 
+  def self.stftime(time, format)
```
jreidinger: I think this is not correct place, it should go to Builtins class and not whole yast namespace, otherwise it confuse anyone as they think it is ruby version of strftime and not our one.

src/ruby/yast/yast.rb
```diff
@@ -197,6 +197,37 @@ def self.#{sname}= (value)
...
+        tm_min:  time.min,
+        tm_hour: time.hour,
+        tm_mday: time.mday,
+        tm_mon:  time.mon - 1, # January is 0 in C++
+        tm_year: time.year - 1900, # Another C++ peculiarity
+        tm_wday: time.wday,
+        tm_yday: time.yday - 1 # January 1st is 0 in C++
+      }
+    rescue NoMethodError
+      raise ArgumentError, "First parameter doesn't look like a time: #{time}"
```
jreidinger: maybe just "Parameter" as it isingle one?

src/binary/Builtin.cc
```diff
@@ -632,6 +632,49 @@ extern "C" {
...
+    res->tm_hour  = tm_hash_value(hash, "tm_hour");
+    res->tm_mday  = tm_hash_value(hash, "tm_mday");
+    res->tm_mon   = tm_hash_value(hash, "tm_mon");
+    res->tm_year  = tm_hash_value(hash, "tm_year");
+    res->tm_wday  = tm_hash_value(hash, "tm_wday");
+    res->tm_yday  = tm_hash_value(hash, "tm_yday");
+    res->tm_isdst = tm_hash_value(hash, "tm_isdst");
+  }
+
+#define STRFTIME_MAX_LENGTH 100
```
jreidinger: any reason why having 100 there? for me usual buffer size is 256 or 512 ( so memory aligned size )

src/binary/Builtin.cc
```diff
@@ -632,6 +632,49 @@ extern "C" {
...
+    struct tm timeinfo;
+
+    Check_Type(format, T_STRING);
+    hash_to_tm(time, &timeinfo);
+
+    if (strftime(res, sizeof(res), RSTRING_PTR(format), &timeinfo)) {
+      return yrb_utf8_str_new(string(res));
+    } else {
+      rb_raise(rb_eRuntimeError, "The result of strftime would be longer than %d characters", STRFTIME_MAX_LENGTH);
+      return Qnil;
```
jreidinger: useless return, rb_raise is usually enough, as it stop execution...but maybe c++ complain then?

src/ruby/yast/builtins.rb
```diff
@@ -674,6 +674,49 @@ def self.y2usernote *args
...
+    def self.stftime(time, format)
+      strftime_wrapper(time_to_tm(time), format)
+    end
+
+    # Creates a hash with the same fields that the C++ time structure.
+    # @private used internally by .strftime
+    # @see http://www.cplusplus.com/reference/ctime/tm/
+    # @param time [Time/DateTime] time to convert
+    # @return [Hash] hash with symbols as keys
+    def self.time_to_tm(time)
```
jreidinger: can you make it private with `private_class_method`
irb session:
```
irb(main):001:0> module M
irb(main):002:1> def self.a
irb(main):003:2> puts "OK"
irb(main):004:2> end
irb(main):005:1> def self.b
irb(main):006:2> puts "should not happen"
irb(main):007:2> end
irb(main):008:1> private_class_method :b
irb(main):009:1> end
=> M
irb(main):010:0> M.a
OK
=> nil
irb(main):011:0> M.b
NoMethodError: private method `b' called for M:Module
	from (irb):11
	from /usr/bin/irb:11:in `<main>'
```

src/binary/Builtin.cc
ancorgs: I don't know. I just followed the examples you pointed out. Both had the same return.

src/ruby/yast/builtins.rb
ancorgs: I know that. But the other methods in the same file also labeled as `@private` are not really declared private. I wanted to keep it consistent.

src/binary/Builtin.cc
ancorgs: No reason. 256 will do it, of course.

src/ruby/yast/builtins.rb
jreidinger: OK, you have my approval to break consistency and do it properly :). It should be both, as I am not sure if yardoc recognize private_class_method and mark it properly.

src/binary/Builtin.cc
```diff
@@ -632,6 +632,61 @@ extern "C" {
...
+    struct tm timeinfo;
+
+    Check_Type(format, T_STRING);
+    hash_to_tm(time, &timeinfo);
+
+    try
+    {
+      // Since std::put_time is not implemented in GCC4.9, we'll use
+      // locale::global and strftime instead of facets
+      // FIXME: I'm not sure about the implications!!!!
```
aschnell: This will effect all C++ libraries linked to YaST. On the other hand setting the locale it is the right thing to do anyway.

src/binary/Builtin.cc
```diff
@@ -632,6 +632,61 @@ extern "C" {
...
+  tm_hash_value(VALUE hash, const char *key)
+  {
+    return NUM2INT(rb_hash_aref(hash, ID2SYM(rb_intern(key))));
+  }
+  
+  // converts a hash to a tm structure
+  // used internally by stftime_wrapper
+  void
+  hash_to_tm(VALUE hash, struct tm *res)
+  {
```
aschnell: It might be a good idea to clear the tm struct in advance using memset. When you look at the definition of tm you can see that is has a few more members thus tools like valgrind might complain about use of uninitialized data when you pass tm to libc functions. The memset call is common practise for ioctl calls.

ancorgs: According to this http://www.ex-parrot.com/~chris/random/initialise.html something like `struct tm timeinfo = {0};` would be more correct than `memset(&timeinfo, 0, sizeof timeinfo);`. @aschnell, does it really make sense in our case? Just curious.

aschnell: I didn't know of that memset problem. If you feel better you can use "= { 0 }" style.

lslezak:
Small <sub>LGTM</sub>  (as I'm not the expert in this area :wink:).

aschnell:
I'm looking forward to have "," instead of "." as decimal separator in German locale ;)

PR #150, opened 2015-07-17
Apply style guide, with Rubocop

mvidner:
I tried hard to separate the autocorrections from manual corrections so that any mistakes and regressions are easy to spot.

BTW I made a tool for this, https://github.com/mvidner/rubocopify


.rubocop.yml
```diff
@@ -0,0 +1,97 @@
...
+Style/Documentation:
+  Enabled: false
+
+# Offense count: 1
+Style/DoubleNegation:
+  Enabled: false
+
+# Offense count: 1
+Style/EachWithObject:
+  Enabled: false
```
jreidinger: this make sense for me to fix

.rubocop.yml
```diff
@@ -0,0 +1,97 @@
...
+
+# Offense count: 1
+Style/EmptyElse:
+  Enabled: false
```
jreidinger: this one also

.rubocop.yml
```diff
@@ -0,0 +1,97 @@
...
+
+# legacy tests
+Style/FileName:
+  Exclude:
+  - tests/ycp/InRuby.rb
+  - tests/ycp/URI.rb
```
jreidinger: it is fine for me to rename test names

src/ruby/yast/builtins.rb
```diff
@@ -117,26 +116,26 @@ def self.isempty object
...
       when ::Array
         res = []
         begin
           object.each do |i|
-            res << block.call(Yast::deep_copy(i))
+            res << block.call(Yast.deep_copy(i))
           end
-        rescue Yast::Break
-          #break skips out of each loop, but allow to keep previous results
+        rescue Yast::Break # rubocop:disable Lint/HandleExceptions
```
jreidinger: ugh, this breaks my internal rule to not add tool specific stuff to code as it decrease readability of code. If needed you can add debug log there.

src/ruby/yast/builtins.rb
```diff
@@ -492,42 +490,42 @@ def self.toset array
 
     # Check if map has a certain key
     # @deprecated use {::Hash#haskey?}
-    def self.haskey map, key
+    def self.haskey(map, key)
       return nil if map.nil? || key.nil?
-      map.has_key? key
+      map.key? key
```
jreidinger: why no parens there?

src/ruby/yast/builtins.rb
```diff
@@ -779,75 +777,75 @@ def self.dpgettext (domain, dirname, text)
...
     end
 
     # Filters characters out of a ::String
     # @deprecated use ruby native method for string handling like {::String#gsub} or {::String#delete}
-    def self.filterchars string, chars
+    def self.filterchars(string, chars)
       return nil if string.nil? || chars.nil?
 
-      return string.gsub(/[^#{Regexp.escape chars}]/, "")
+      string.gsub(/[^#{Regexp.escape chars}]/, "")
```
jreidinger: also parens missing here

src/ruby/yast/logger.rb
```diff
@@ -15,47 +15,52 @@ def y2_logger_helper(level,args)
...
     if backtrace
       y2_logger_helper(level, [2, "------------- Backtrace begin -------------"])
-      caller(3).each {|frame| y2_logger_helper(level, [4, frame])}
+      caller(3).each { |frame| y2_logger_helper(level, [4, frame]) }
       y2_logger_helper(level, [2, "------------- Backtrace end ---------------"])
     end
   end
 
   module_function
+
```
jreidinger: ugh, why??? I really want to have it as tight as possible to y2debug as it modify accessibility of method. maybe even better to use `module_function def y2debug(*args)` ?

src/ruby/yast/path.rb
```diff
@@ -21,56 +21,57 @@ def clone
...
+        other_component = other.send(:components)[i].sub(/\A"(.*)"\Z/, "\\1")
         res = our_component <=> other_component
         return res if res != 0
       end
-      return size <=> other.size
+      size <=> other.size
     end
 
-  private
+    private
```
jreidinger: I do not like this change :) I prefer to have method visibility section change on "class" level indentation. Rationale is that many methods just modify next method. But this one modify all remaining definitions unless modified.

src/ruby/yast/builtins.rb
mvidner: Apparently Rubocop does not care about parens at the *call* point, only at the *definition* point.

src/ruby/yast/path.rb
mvidner: I see you have codified it in yast2: https://github.com/yast/yast-yast2/blob/06efb8a3f824b7d53873192a0d11e9fbf611e8b5/.rubocop.yml#L33
I don't really care about this particular case. Should we adapt https://github.com/yast/yast-devtools/blob/master/ytools/y2tool/rubocop_yast_style.yml ? How does it play with https://github.com/SUSE/style-guides/blob/master/Ruby.md anyway?

shundhammer:
Looks good to me.

But then, after a few thousand lines, the whitespace diffs began to blur before my eyes... ;-)

src/ruby/yast/path.rb
jreidinger: I think we should adapt yast style guide. I play quite good with suse style guide, as it is not mentioned there explicitelly

PR #152, opened 2015-09-15
another attempt to fix terminated string problem (bnc#945299)

jreidinger:


src/binary/Y2RubyReference.cc
```diff
@@ -9,10 +9,14 @@ YCPValue ClientFunction::evaluateCall()
...
   {
     VALUE value = ycpvalue_2_rbvalue(m_call.value(i));
-    RB_GC_GUARD(value);
+    rb_gc_register_address(&value);
     params[i] = value;
   }
   YCPValue res = rbvalue_2_ycpvalue(rb_funcall3(object, rb_intern("call"),m_call.size(), params));
+  for (int i = 0; i < m_call.size(); ++i)
+  {
+    rb_gc_unregister_address(params+i);
```
mvidner: I think this may not work, you are unregistering a different pointer than you registered. The VALUE pointed to is the same, but I guess the unregistration will silently fail.

IMHO it is better to use `rb_gc_..._address(params + i)` in both cases.

src/binary/YRuby.cc
```diff
@@ -190,7 +190,7 @@ YCPValue YRuby::callInner (string module_name, string function,
     YCPValue v = argList->value(i);
     y2debug("Adding argument %d of type %s", i, v->valuetype_str());
     VALUE vr = ycpvalue_2_rbvalue(v);
-    RB_GC_GUARD(vr);
+    rb_gc_register_address(&vr);
```
mvidner: same here

src/binary/Y2RubyReference.cc
jreidinger: maybe not silently fail, customer report segfault, in different unregister for referenced value, so maybe it is problem. I will fix it on monday and create new patch

lslezak:
The patch seems to be OK according to the reporter, we should fix it also in SLES12-SP1...

mvidner:
LGTM! Thanks for the debugging!

 I give you the changelog, you bump the version :)
> Used rb_gc_register_address to fix 'method to_s called on terminated object' during package installation (bsc#945299).

ancorgs:
This is my first contact with the C part of the bindings so I only have a noob question. Is the difference between the usage of `RB_GC_GUARD` and `rb_gc_(un)register_address` obvious enough to somebody who knows the background? In other words, would be useful a comment somewhere to avoid a similar bug in the future?

jreidinger:
@ancorgs RB_GC_GUARD make variable volative, so gcc cannot optimize it out. rb_gc_(un)register_address register(un-) from garbage collector. So it is stronger, as garbage collector cannot unregister. Basically even previous stuff should be enough as it is passed as parameter to method, so from my POV ruby should not destruct it, but it somehow ignore it.

ancorgs:
@jreidinger That was exactly my point. If the behavior was unexpected for you, then it deserves a comment somewhere in the source code.

jreidinger:
@ancorgs question is where as it is used in multiple places

ancorgs:
@jreidinger I don't know. But better everywhere than nowhere :-)

ancorgs:
LGTM, thanks

PR #153, opened 2015-09-16
Test for bnc#943757, fixed in libyui-ncurses 2.47.3

ancorgs:
I didn't manage to write a test reproducing the bug in libyui-test, so I wrote a test here (which, on the other hand, fits better the bug description, since YaST and Ruby are involved).

Should we add ```BuildRequires: libyui-ncurses >= 2.47.3``` to the spec file? I find it a little bit overkill, but I'm open to suggestions.

This can be merged after https://github.com/libyui/libyui-ncurses/pull/40 or https://github.com/libyui/libyui-ncurses/pull/41 (whichever wins the contest), since both make the test happy.

### DIFF
--- a/tests/ruby/std_streams_spec.rb
+++ b/tests/ruby/std_streams_spec.rb
@@ -1,42 +1,40 @@
 #! /usr/bin/env rspec
 
+# We do not have a proper ncurses in travis at the moment
+exit 0 if ENV["TRAVIS"]
+
 require_relative "test_helper"
 require "yast/ui_shortcuts"
 
 Yast.import "UI"
 
-# We do not have a proper ncurses in travis at the moment
-if !ENV["TRAVIS"]
-  def std_puts(message)
-    $stdout.puts "stdout: #{message}"
-    $stderr.puts "stderr: #{message}"
+def std_puts(message)
+  $stdout.puts "stdout: #{message}"
+  $stderr.puts "stderr: #{message}"
+end
+
+# Regression test for the fix of bnc#943757 implemented
+# in libyui-ncurses 2.47.3
+describe "streams redirection in libyui-ncurses" do
+  include Yast::UIShortcuts
+
+  around do |example|
+    Yast.ui_component = "ncurses"
+    Yast::UI.OpenUI
+    example.run
+    Yast::UI.CloseUI
+
+    # Having an expectation in the around block looks weird, but using around
+    # to execute OpenUI/CloseUI was needed to make the bug pop up.
+    #
+    # In addition to not crashing, these messages should be displayed when
+    # running RSpec, not sure if it's possible to check that.
+    expect { std_puts "tty is free again" }.to_not raise_error
   end
 
-  # Regression test for the fix of bnc#943757 implemented
-  # in libyui-ncurses 2.47.3
-  describe "streams redirection in libyui-ncurses" do
-    include Yast::UIShortcuts
-
-    before do
-      Yast.ui_component = "ncurses"
-      Yast::UI.OpenUI
-    end
-
-    after do
-      Yast::UI.CloseUI
-
-      # Having an expectation in the after block looks weird, but using
-      # before/after to execute OpenUI/CloseUI was needed to make the bug popup
-      #
-      # In addition to not crashing, these messages should be displayed when
-      # running RSpec, not sure if it's possible to check that
-      expect { std_puts "tty is free again" }.to_not raise_error
-    end
-
-    it "does not fall apart when stderr is used" do
-      Yast::UI.OpenDialog(PushButton("Hello, World!"))
-      expect { std_puts "NCurses is using the tty" }.to_not raise_error
-      Yast::UI.CloseDialog
-    end
+  it "does not fall apart when stderr is used" do
+    Yast::UI.OpenDialog(PushButton("Hello, World!"))
+    expect { std_puts "NCurses is using the tty" }.to_not raise_error
+    Yast::UI.CloseDialog
   end
 end
--- a/tests/ruby/std_streams_spec.rb
+++ b/tests/ruby/std_streams_spec.rb
@@ -30,12 +30,12 @@ if !ENV["TRAVIS"]
       #
       # In addition to not crashing, these messages should be displayed when
       # running RSpec, not sure if it's possible to check that
-      expect {std_puts "tty is free again"}.to_not raise_error
+      expect { std_puts "tty is free again" }.to_not raise_error
     end
 
     it "does not fall apart when stderr is used" do
       Yast::UI.OpenDialog(PushButton("Hello, World!"))
-      expect {std_puts "NCurses is using the tty"}.to_not raise_error
+      expect { std_puts "NCurses is using the tty" }.to_not raise_error
       Yast::UI.CloseDialog
     end
   end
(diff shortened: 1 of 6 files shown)


tests/ruby/std_streams_spec.rb
```diff
@@ -0,0 +1,42 @@
...
+    after do
+      Yast::UI.CloseUI
+
+      # Having an expectation in the after block looks weird, but using
+      # before/after to execute OpenUI/CloseUI was needed to make the bug popup
+      #
+      # In addition to not crashing, these messages should be displayed when
+      # running RSpec, not sure if it's possible to check that
+      expect {std_puts "tty is free again"}.to_not raise_error
+    end
```
jreidinger: why not using `around`?

tests/ruby/std_streams_spec.rb
```diff
@@ -0,0 +1,42 @@
+#! /usr/bin/env rspec
+
+require_relative "test_helper"
+require "yast/ui_shortcuts"
+
+Yast.import "UI"
+
+# We do not have a proper ncurses in travis at the moment
+if !ENV["TRAVIS"]
```
jreidinger: I do not like much complete enclosing whole file. What about `exit 0 if ENV["TRAVIS"]` ?

jreidinger:
@ancorgs regarding BuildRequire ... if it is needed for build add it, otherwise do not add it :)

ancorgs:
Well, it's required for this test to succeed.

jreidinger:
@ancorgs we run test during build, so it is needed then, not?

ancorgs:
Hum, I think there is another problem. I would not expect the whole thing to run properly in jenkins (and probably other headless systems) unless we use `libyui-terminal rspec whatever` (it has been needed in the past for YaST tests opening dialogs). But using `libyui-terminal` actually hides the bug and the test succeeds even with the broken libyui.

@mvidner Ideas?

We have to wait until libyui-ncurses 2.47.3 is available before merging anyways

jreidinger:
then maybe use instead env travis check for headless systems?

ancorgs:
Well, the new libyui-ncurses is already in Tumbleweed and SP1, so we have to decide what to do about this.

1) Simply skip the execution in travis and jenkins? In that case, what would be a sensible check to replace the current `ENV['travis']`?

2) Make it run with `libyui-terminal` by changing the `check` section of the rpm? That will hide the error when running in jenkins, but the error (in case of regression) should still be visible when doing a simple `rake test:unit`.

Simply merge and let's see if it runs as it is in jenkins, without `libyui-terminal`? I would expect some 3) fireworks, based on previous experiences. But...

@mvidner What do you say?

mvidner:
@ancorgs As you said, `libyui-terminal` hides the bug because it saves the command's output by redirecting stdio from the console to a file. I think it is worth to make a special case for this test and give it its own runner that uses `screen` without redirection.

BTW `rake test:unit` does nothing for this package because it expects a `test` subdir where we have `tests`.

As for the additional dependency on libyui-ncurses, I don't like that very much. I think that for an integration test like this we should add a new spec file, yast2-ruby-bindings-test.spec which BuildRequires libyui-ncurses.

Detecting Jenkins... how about `Etc.getlogin == "jenkins"` ? Or setting up our instance to set CI=true
CONTINUOUS_INTEGRATION=true like [Travis does](http://docs.travis-ci.com/user/environment-variables/#Default-Environment-Variables).

jreidinger:
@mvidner tests subdir is intention as this package do not work easily ( it require at first built C bindings before it can run and even then it is quite tricky to set correctly paths, so it depends on cmake to run tests.

mvidner:
After sorting out a mixture of ruby2.1+2.2 on my VM I finally got to run the test myself.

`make test` will not trigger the bug because CMake redirects the output to `./Testing/Temporary/LastTest.log`

I have tested that this addition to the `Rakefile` catches the bug if the fix is not present:
```ruby
desc "Runs all tests"
task :test => ["test:unit", "test:integration"]

# Redefine a generic task for this package
task "test:unit" do
  cd "build" do
    sh "make test"
  end
end

desc "Runs integration tests"
task "test:integration" do
  cd "build" do
    f = "../tests/ruby/std_streams_spec.rb"
    cmd = "rspec #{f}"
    sh "screen -D -m sh -c '#{cmd}; echo \$? > /tmp/exit'"
    if File.read("/tmp/exit") != "0\n"
        fail "Test failed: '#{cmd}'. Rerun manually to see the cause."
    end
  end
end
```

ancorgs:
Ok, I will add that to the Rakefile if nobody is against. Looks like a nice compromise.

mvidner:
According to https://ci.opensuse.org/env-vars.html/ , `JENKINS_HOME` looks like a good variable to check, if we still need that.

jreidinger:
@mvidner @ancorgs I do not like that change in Rakefile. Reason is simple. It is common that you can do `git clone` and `rake test:unit` which do not work in this case as it have hardcoded build directory for cmake and also missing C part building step, so it failed

ancorgs:
@jreidinger right. I have just checked and found that problem

ancorgs:
@jreidinger @mvidner Fixed to make sure the test runs using screen during package building.

The split in two packages suggested by @mvidner is still pending. Feel free to commit it yourself, since I'm not sure about the details (I have granted you permissions on my fork, so you can directly add commits to this PR).

mvidner:
From the OBS viewpoint, the package split is as easy as comitting another spec file. OBS wil automatically create a package named after it and link it to the original. Then we only need to watch for the build failures.

mvidner:
I'll make the split

mvidner:
@ancorgs @jreidinger after my changes it looks good to myself :) so please check.
I tested with `rake osc:build[yast2-ruby-bindings-test.spec]`

ancorgs:
Looks good to me, but I would wait for @jreidinger's blessing before merging.

jreidinger:
I have few concerns regarding splitting:

1) jenkins runs rake osc:sr which as dependency run rake osc:build, so for me it looks like it is not automatic run
2) adding new package in this phase of SP1 and leap release maybe not succeed
3) splitted spec provide more complexity, need also second changelog and make things more tricky, do you measure benefit of such changes?

mvidner:
1) yes, this would need a separate jenkins job
2) IMHO not a problem to add a source package without adding the resulting (empty) binary package
3) we already have yast packages with multiple spec/changes, eg.
./yast/yast-pkg-bindings/package/yast2-pkg-bindings-devel-doc.spec
./yast/yast-pkg-bindings/package/yast2-pkg-bindings.spec
so `rake version:bump` deals with that.

Anyway, I can back off the split and we add a note for the OBS guardians in the SR that we are growing the dependency chain and if that is a problem they should let us know and we will split it.

mvidner:
My commits after I said "I'll make the split" are now in https://github.com/mvidner/yast-ruby-bindings/commits/split-test-subpackage and I have force pushed this branch not to include them.

jreidinger:
ok, LGTM

mvidner:
> and we add a note for the OBS guardians in the SR that 

Ugh, I forgot that we haven't bumped the version or even mentioned it in .changes (commit https://github.com/mvidner/yast-ruby-bindings/commit/9b743cef9ed585a9831516cc876147e7cb454902 was reverted) so there is no SR to add a note to.

ancorgs:
Oops. Done in https://github.com/yast/yast-ruby-bindings/pull/154

PR #157, opened 2015-12-01
Revert cell

jreidinger:
for more details see https://bugzilla.suse.com/show_bug.cgi?id=956380

### DIFF
--- a/package/yast2-ruby-bindings.changes
+++ b/package/yast2-ruby-bindings.changes
@@ -1,4 +1,11 @@
 -------------------------------------------------------------------
+Tue Dec  1 16:06:11 UTC 2015 - jreidinger@suse.com
+
+- revert last change as it break using cell for quering table cell
+  values (bnc#956380)
+- 3.1.42
+
+-------------------------------------------------------------------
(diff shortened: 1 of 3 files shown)


aschnell:
LGTM, but as you can see we need test cases and better documentation.

jreidinger:
@aschnell agreed, I add to scrum board documentation for all UI elements as we really need to document it and not still using old documentation, that is no longer maintained.

package/yast2-ruby-bindings.changes
```diff
@@ -1,4 +1,11 @@
 -------------------------------------------------------------------
+Tue Dec  1 16:06:11 UTC 2015 - jreidinger@suse.com
+
+- revert last change as it break using cell for quering table cell
```
mvidner: > Reverted the last change because it broke updating the table in Service Manager
> (because the distinction between :cell and :Cell was lost; bnc#956380)

mvidner:
Thanks!
I just want to reiterate: this illustrates why it does make sense to write real bug numbers to the changes file. A real bug means someone had to explain to a rubber duck what the problem is and what the fix is.

mvidner:
LGTM

PR #158, opened 2016-01-11
Do not crash when logging an invalid UTF-8 string (bsc#932014)

lslezak:
- 3.1.43

Fixes https://bugzilla.suse.com/show_bug.cgi?id=932014

### DIFF
--- a/src/ruby/yast/logger.rb
+++ b/src/ruby/yast/logger.rb
@@ -1,6 +1,10 @@
 require "yastx"
 require "yast/builtins"
 
+# FIXME: a workaround for https://github.com/yast/yast-ruby-bindings/issues/133
+require "enc/encdb.so"
+require "enc/trans/transdb.so"
+
 module Yast
   # @private
   module_function def y2_logger_helper(level, args)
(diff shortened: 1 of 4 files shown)


src/ruby/yast/logger.rb
```diff
@@ -1,6 +1,10 @@
 require "yastx"
 require "yast/builtins"
 
+# FIXME: a workaround for https://github.com/yast/yast-ruby-bindings/issues/133
+require "enc/encdb.so"
+require "enc/trans/transdb.so"
```
jreidinger: I think that this workaround should be placed near ruby initializer: https://github.com/yast/yast-ruby-bindings/blob/master/src/binary/YRuby.cc#L75

lslezak: Interesting, I put

```c
  rb_require("enc/encdb.so");
  rb_require("enc/trans/transdb.so");
```

there and it did not help and raise the `code converter not found (ASCII-8BIT to UTF-8)`... :question:
(I tried it also without the `.so` suffix, no change...)

jreidinger: interesting, especially if you see that we already have there encdb hack...do you place it after or before?

lslezak: Um, I played with it a bit but I did not succeed... :worried: 

I tried different places for the `require` calls, also tried using `Init_enc()` [as mentioned here](http://stackoverflow.com/questions/25449061/how-do-you-fully-initialize-an-embedded-ruby-vm-in-a-c-application), tried several places, but with the same result - the exception was still raised.

Only when placing the `require`s in the Ruby file worked.

I do not want to spend more time here, let's use this workaround and improve it later when really needed....

jreidinger:
LGTM

src/ruby/yast/logger.rb
```diff
@@ -17,7 +21,20 @@ module Yast
       end
     end
 
-    res = Builtins.sformat(*args)
+    # replace invalid characters by the replacement symbol
+    # see https://en.wikipedia.org/wiki/Specials_(Unicode_block)#Replacement_character
+    safe_args = args.map do |arg|
+      return arg unless arg.is_a?(::String)
```
jreidinger: This is root of issue why now majority of yast modules failed. Just try it:
```
def a
  res = [0, 1].map do |a|
    return a
  end

  puts "OK"

  res
end

puts a.inspect
```

kobliha: And that's also why we should have a unit test for this

PR #159, opened 2016-01-12
Fix Encoding name conflict

lslezak:
Fixes conflict between `Yast::Encoding` and `::Encoding` (from std Ruby).


jreidinger:
LGTM

PR #160, opened 2016-01-13
Fix return

jreidinger:


lslezak:
LGTM, thanks for the fix! :+1:

PR #162, opened 2016-01-13
Enc init

jreidinger:


lslezak:
Tested and works correctly. LGTM.

Thanks!

PR #163, opened 2016-01-20
A detailed docs for Builtins.remove.

mvidner:


src/ruby/yast/builtins.rb
```diff
@@ -146,9 +146,18 @@ def self.maplist(object, &block)
...
-    # - Remove item from term
+    # Remove *element* from a **copy** of *object*.
+    #
+    # @param object [::Hash,::Array,Yast::Term,nil] a container
+    # @param element [Object,Integer,nil]
+    #   Any key for Hash;
+    #   Integer for Array and Term;
+    #   terms are indexed from 1 (**one**)
+    # @return [Object,nil]
+    #   Returns *object* unchanged if either *object* or *element* is `nil`;
```
jreidinger: this is not true as it returns copy if element is nil. So not unchanged. For nil object it it not interesting as nil is immutable.

src/ruby/yast/builtins.rb
```diff
@@ -146,9 +146,18 @@ def self.maplist(object, &block)
...
 
-    # - Removes element from a list
-    # - Remove key/value pair from a map
-    # - Remove item from term
+    # Remove *element* from a **copy** of *object*.
+    #
+    # @param object [::Hash,::Array,Yast::Term,nil] a container
+    # @param element [Object,Integer,nil]
+    #   Any key for Hash;
+    #   Integer for Array and Term;
```
jreidinger: maybe document that for negative number it returns copy of container without modification, so nothing like -1 is last or like that

src/ruby/yast/builtins.rb
mvidner: Ah! Thanks for the catch.

mvidner:
Updated.

jreidinger:
LGTM

PR #164, opened 2016-02-09
Documented Builtins.regexpsub

mvidner:
added more tests for it

### DIFF
--- a/src/ruby/yast/builtins.rb
+++ b/src/ruby/yast/builtins.rb
@@ -1044,6 +1044,29 @@ module Yast
     #
     # In a condition, use `string =~ pattern` which returns integer or nil.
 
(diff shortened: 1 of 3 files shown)


lslezak:
LGTM

src/ruby/yast/builtins.rb
```diff
@@ -1044,6 +1044,29 @@ def self.toupper(string)
     #
     # In a condition, use `string =~ pattern` which returns integer or nil.
 
+    # @method self.regexpsub(input, pattern, output)
+    #
+    # @param input       [String] a string to search
+    # @param pattern     [String] a regex in the C(!) syntax
```
jreidinger: it is actually not C syntax but POSIX syntax. C do not have any regexp syntax

PR #166, opened 2016-03-07
adapt to updated rubocop and remove some exceptions

jreidinger:


mvidner:
Please add a changes entry, just in case we introduce a bug here.

src/ruby/yast/builtins.rb
```diff
@@ -366,13 +366,11 @@ def self.swap(list, offset1, offset2)
...
-          res.concat list[0..offset1 - 1]
-        end
-        res.concat list[offset1..offset2].reverse!
-        if offset2 < list.size - 1
-          res.concat list[offset2 + 1..-1]
-        end
+
+        res.concat(list[0..offset1 - 1]) if offset1 > 0
+        res.concat(list[offset1..offset2].reverse!)
+        res.concat(list[offset2 + 1..-1]) if offset2 < list.size - 1
```
mvidner: Builtins::List.swap is used only twice in a single file; we could remove it:
https://github.com/yast/yast-storage/blob/3e2a3edaa59c1e6d7c8805d2e82e017ea496eebd/src/modules/DualMultiSelectionBox.rb#L604

jreidinger: how we could remove it if it is used? I do not get it.

jreidinger:
ok, non-trivial change, so I will add changelog entry

src/ruby/yast/builtins.rb
mvidner: I mean, we could rewrite the places where it it used and THEN remove it :)

mvidner:
LGTM

PR #168, opened 2016-05-30
Improve the debugger support (FATE#318421)

lslezak:
Use the same code also at run time, allow using `Y2DEBUGGER` also in installed system.

Now you can run `Y2DEBUGGER=1 yast2 <client>` and the debugger window will automatically pop up.

- 3.1.48

### DIFF
--- a/src/ruby/yast/debugger.rb
+++ b/src/ruby/yast/debugger.rb
@@ -62,14 +62,16 @@ module Yast
         # https://github.com/deivid-rodriguez/byebug/blob/master/GUIDE.md
       end
 
-      # start the Ruby debugger if "Y2DEBUGGER" or "y2debugger" environment
-      # variable is set to "1", "remote" or "manual"
+      # start the Ruby debugger if "Y2DEBUGGER" environment
+      # variable is set to "1", "remote" or "manual" (the test is case
+      # insensitive, "y2debugger" variable can be also used)
       def start_from_env
-        # do not start the debugger again for each client started, run it only
-        # once for the very first client
-        return if @debugger_started
+        # do not evaluate the debugger request again for each client started,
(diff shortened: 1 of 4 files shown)


src/ruby/yast/wfm.rb
```diff
@@ -247,5 +252,29 @@ def self.run_client(client)
...
+      # initialization, if it is initialized inside the Debugger module the
+      # ncurses UI segfaults :-(
+      # interestengly, the Qt UI works correctly...
+      Yast.import "UI"
+      log.info "text mode: #{UI.TextMode}"
+
+      log.info "Debugger set to: #{debug}"
+      Debugger.start(remote: debug == "remote", start_client: debug != "manual")
+      @debugger_started = true
+    end
```
jreidinger: well, for me this is debugger code, so I expect it in debugger file and just calling from here like `Debugger.start_if_requested`

jreidinger:
LGTM...btw running it in WFM allows interesting possibility to start debugger only for specific client.

src/ruby/yast/debugger.rb
```diff
@@ -73,6 +95,17 @@ def installed?
 
     private
 
+      # read the debugger value from Y2DEBUGGER environment variable,
+      # do case insensitive match
+      # @return [String,nil] environment value or nil if not defined
+      def read_dbg_env
```
jreidinger: I think simple `env_value` is enough for this method. Read connect in my mind with something expensive, which is not this case.

jreidinger:
just NP, so LGTM

lslezak:
Evaluating `ENV.keys.sort.find` is quite expensive, I have modified it so it's evaluated only once, not for every client started.

jreidinger:
LGTM

PR #169, opened 2016-06-28
Profiler

jreidinger:
caller optimization speed up minimal installation by 3% as it affect all logging.

Any idea where to document new y2profiler variable beside SDB:linuxrc wiki page?

### DIFF
--- a/src/binary/YRuby.cc
+++ b/src/binary/YRuby.cc
@@ -110,6 +110,9 @@ YRuby::~YRuby()
     y2milestone( "Shutting down ruby interpreter." );
     //ruby_finalize(); Do not finalize to allow clear work inside ruby
     _y_ruby_finalized = true;
+    // call at_exit methods, it is part of ruby_finalize, which we have to skip,
+    // so use only needed parts
+    rb_exec_end_proc();
 }
 
 
--- a/src/ruby/yast/profiler.rb
+++ b/src/ruby/yast/profiler.rb
@@ -1,0 +1,58 @@
+
+require "yast"
+
+module Yast
+  class Profiler
+    class << self
+      include Yast::Logger
+
+      RESULT_PATH = "/tmp/yast_profile.txt"
+
+      # Start the Ruby Profiler. It start profilling. It also disables ruby VM
+      # optimizations, so code execution will be slower.
+      def start
+        raise "multiple profiller start detected" if @started
+        @original_compile_options = RubyVM::InstructionSequence.compile_option
+        @started = true
+        require "profiler"
+
+        # turn on tracing and turn of specialized instruction to get complete profilling
+        RubyVM::InstructionSequence.compile_option = {
+          trace_instruction:       true,
+          specialized_instruction: false
+        }
+
+        at_exit { File.open(RESULT_PATH, "w") { |f| Profiler__.print_profile(f) } }
+        Profiler__.start_profile
+      end
+
+      # start the Ruby profiler if "Y2PROFILER" environment
+      # variable is set to "1" or "true"(the test is case
+      # insensitive, "y2profiler" variable can be also used)
+      def start_from_env
+        # do not evaluate again for each client started, run the evaluation only once
+        return if @profiler_handled
+        @profiler_handled = true
+
+        return unless env_value
+
+        log.info "profiler enabled"
+        start
+      end
+
+    private
+
+      # read the Y2PROFILER environment variable,
+      # do case insensitive match
+      # @return [Boolean] true if enabled
+      def env_value
+        # sort the keys to have a deterministic behavior and to prefer Y2DEBUGGER
+        # over the other variants, then do a case insensitive search
+        key = ENV.keys.sort.find { |k| k.match(/\AY2PROFILER\z/i) }
+        return false unless key
+
+        ["1", "true"].include?(ENV[key])
+      end
+    end
+  end
+end
--- a/src/ruby/yast/profiler.rb
+++ b/src/ruby/yast/profiler.rb
@@ -6,7 +6,7 @@ module Yast
     class << self
       include Yast::Logger
 
-      RESULT_PATH = "/tmp/yast_profile.txt"
+      RESULT_PATH = "/var/log/YaST2/profiler_result.txt"
 
       # Start the Ruby Profiler. It start profilling. It also disables ruby VM
       # optimizations, so code execution will be slower.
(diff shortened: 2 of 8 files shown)


jreidinger:
package not dumped, so not ready to merge

mvidner:
package definitely not dumped!
oh, you mean the version has not been bumped :)

jreidinger:
yeah, version :)

src/ruby/yast/profiler.rb
```diff
@@ -0,0 +1,58 @@
+
+require "yast"
+
+module Yast
+  class Profiler
+    class << self
+      include Yast::Logger
+
+      RESULT_PATH = "/tmp/yast_profile.txt"
```
mvidner: this is open to symlink attacks

src/ruby/yast/profiler.rb
```diff
@@ -0,0 +1,58 @@
...
+
+      # Start the Ruby Profiler. It start profilling. It also disables ruby VM
+      # optimizations, so code execution will be slower.
+      def start
+        raise "multiple profiller start detected" if @started
+        @original_compile_options = RubyVM::InstructionSequence.compile_option
```
mvidner: this would be only useful if we had a `stop_profiler`

src/ruby/yast/profiler.rb
```diff
@@ -0,0 +1,58 @@
...
+        start
+      end
+
+    private
+
+      # read the Y2PROFILER environment variable,
+      # do case insensitive match
+      # @return [Boolean] true if enabled
+      def env_value
+        # sort the keys to have a deterministic behavior and to prefer Y2DEBUGGER
```
lslezak: It's actually `Y2PROFILER` :wink:

src/ruby/yast/profiler.rb
```diff
@@ -0,0 +1,58 @@
...
+
+      # Start the Ruby Profiler. It start profilling. It also disables ruby VM
+      # optimizations, so code execution will be slower.
+      def start
+        raise "multiple profiller start detected" if @started
+        @original_compile_options = RubyVM::InstructionSequence.compile_option
+        @started = true
+        require "profiler"
+
+        # turn on tracing and turn of specialized instruction to get complete profilling
```
lslezak: Typo: `turn off`

BTW what is a *specialized instruction*? A short description or a docu link would be nice here.

jreidinger: It is specially optimized common instructions, it is described in book "Ruby Under a Microscope: Learning Ruby Internals Through Experiment" I am not sure if it is described somewhere in publicly available web page. So it basically turn off VM optimization which will hide some ruby methods.

jreidinger:
updated

src/ruby/yast/profiler.rb
```diff
@@ -0,0 +1,71 @@
...
+        RubyVM::InstructionSequence.compile_option = {
+          trace_instruction:       true,
+          specialized_instruction: false
+        }
+
+        at_exit { stop }
+        Profiler__.start_profile
+      end
+
+      # stops profiling. If f is nil, then store profiling into RESULT_PATH file.
```
mvidner:
```rb
# Stop profiling
# param f [IO] an IO stream to print the profile to; if nil, uses a file at RESULT_PATH
```

src/ruby/yast/profiler.rb
```diff
@@ -0,0 +1,71 @@
...
+      # read the Y2PROFILER environment variable,
+      # do case insensitive match
+      # @return [Boolean] true if enabled
+      def env_value
+        # sort the keys to have a deterministic behavior and to prefer Y2PROFILER
+        # over the other variants, then do a case insensitive search
+        key = ENV.keys.sort.find { |k| k.match(/\AY2PROFILER\z/i) }
+        return false unless key
+
+        ["1", "true"].include?(ENV[key])
```
mvidner: I think we should allow only "1" to be consistent with the similar flags already used in YaST and Linuxrc. Otherwise also "yes" and "on" would be candidates and you would not remember what is allowed where.

src/ruby/yast/yast.rb
```diff
@@ -183,7 +183,7 @@ def self.import(mname)
       if (stype == :function)
         m.module_eval <<-"END"
           def self.#{sname}(*args)
-            caller[0].match BACKTRACE_REGEXP
+            caller(1,1).first.match BACKTRACE_REGEXP
```
mvidner: Aaah! Here `caller` was used for all calls across modules, not only logging.

src/ruby/yast/profiler.rb
```diff
@@ -0,0 +1,71 @@
+
+require "yast"
+
+module Yast
+  class Profiler
+    class << self
+      include Yast::Logger
+
+      RESULT_PATH = "/var/log/YaST2/profiler_result.txt"
```
lslezak: NP: That target is writable only for *root*, if you have an Yast module which can run without *root* privileges than you cannot profile it.

My suggestions:

- Use a temporary file with `yast2-profile-` prefix, it ensures the file name is unique.
- Ask the user for the file name using [UI::AskForSaveFileName()](https://doc.opensuse.org/projects/YaST/SLES11/tdg/AskForSaveFileName.html)
- Pass the file name via environment, e.g. `Y2PROFILERFILE` (if not set then use one of the options above)

src/ruby/yast/yast.rb
jreidinger: only for calls to component system like Pkg, UI or perl ones.

src/ruby/yast/profiler.rb
jreidinger: I do not like asking as it can affect profilling result. Actually it is developer tool, so I don't see as such big problem to profile only with root privileges

lslezak:
OK, LGTM

src/ruby/yast/yast.rb
jreidinger: and it was manually tested by installation with new ruby bindings. I expect it will be quickly recognized, if it is wrong.

mvidner:
LGTM

(This PR took 1.13 days to LGTM)

src/binary/YRuby.cc
```diff
@@ -110,6 +110,9 @@ YRuby::~YRuby()
     y2milestone( "Shutting down ruby interpreter." );
     //ruby_finalize(); Do not finalize to allow clear work inside ruby
     _y_ruby_finalized = true;
+    // call at_exit methods, it is part of ruby_finalize, which we have to skip,
+    // so use only needed parts
+    rb_exec_end_proc();
```
mvidner: One of the crashes caused by this: https://ci.opensuse.org/job/yast-yast2-github-push/119//console
@jreidinger so this runs only at the time global objects are destroyed which is too late :(

PR #170, opened 2016-06-28
[WIP] Build ruby-lint definitions

mvidner:
Part of https://hackweek.suse.com/projects/1580

ruby-lint relies on pre-generated [definition files](http://code.yorickpeterse.com/ruby-lint/latest/file.definitions.html) to check APIs that are called in analyzed code. 

This PR builds the definitions for this package so that they can be used when linting yast-yast2 etc.


mvidner:
ruby-lint is still too broken to be deployed. If I manage to fix it (or a subset of it) then I revive this PR.

PR #171, opened 2016-06-30
Fix ruby terminating

jreidinger:


src/binary/CMakeLists.txt
```diff
@@ -99,7 +99,10 @@ target_link_libraries( builtinx ${YAST_YCP_LIBRARY} )
 target_link_libraries( builtinx ${YAST_PLUGIN_SCR_LIBRARY} )
 target_link_libraries( builtinx ${YAST_PLUGIN_WFM_LIBRARY} )
 target_link_libraries( builtinx ${YAST_PLUGIN_UI_LIBRARY} )
-target_link_libraries( builtinx crypt )
+find_library( CRYPT_LIBRARY crypt )
+if ( CRYPT_LIBRARY )
+  target_link_libraries( builtinx ${CRYPT_LIBRARY} )
+endif()
```
mvidner: wouldn’t it be better to fail if the library is not found?

mvidner:
The code looks good. Have you tested it with the profiler enabled?

jreidinger:
Now tested and works as expected. Without any issues.

src/binary/CMakeLists.txt
jreidinger: to be honest, I try to touch cmake as less as possible. So I just tried to fix warning. I can revert it.

mvidner:
LGTM

PR #172, opened 2016-07-14
rubocop update

jreidinger:


.rubocop.yml
```diff
@@ -68,7 +72,12 @@ Style/SingleLineBlockParams:
 Style/TrivialAccessors:
   Enabled: false
 
+
```
lslezak: NP: looks like an extra empty line...

src/ruby/yast/builtins.rb
```diff
@@ -350,7 +346,7 @@ def self.flatten(value)
 
       value.reduce([]) do |acc, i|
         return nil if i.nil?
-        acc.push(*Yast.deep_copy(i))
+        acc.concat(Yast.deep_copy(i))
```
lslezak: Oh, nice catch! :+1:

jreidinger: it is actually rubocop auto fix :)

lslezak:
LGTM

PR #173, opened 2016-09-14
Rescue "invalid byte sequence in UTF-8", with a custom message

mvidner:
[bsc#992821](https://bugzilla.suse.com/show_bug.cgi?id=992821)

### DIFF
--- a/src/ruby/yast/wfm.rb
+++ b/src/ruby/yast/wfm.rb
@@ -199,7 +199,7 @@ module Yast
       if e.is_a?(ArgumentError) && e.message =~ /invalid byte sequence in UTF-8/
         msg += "A string was encountered that is not valid in UTF-8.\n" \
                "The system encoding is #{Encoding.locale_charmap.inspect}.\n" \
-               "This is a known bug bsc#992821.\n\n"
+               "Refer to https://www.suse.com/support/kb/doc?id=7018056.\n\n"
       end
 
       msg + "Details: #{e.message}\n" \
--- a/src/ruby/yast/wfm.rb
+++ b/src/ruby/yast/wfm.rb
@@ -194,9 +194,16 @@ module Yast
     # @param [Exception] e the caught exception
     # @return [String] human readable exception description
     def self.internal_error_msg(e)
-      "Internal error. Please report a bug report with logs.\n" \
-        "Details: #{e.message}\n" \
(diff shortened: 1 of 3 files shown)


mvidner:
- [ ] fill in the number
- [ ] version + changelog

src/ruby/yast/wfm.rb
```diff
@@ -194,9 +194,17 @@ def self.ask_to_run_debugger?
     # @param [Exception] e the caught exception
     # @return [String] human readable exception description
     def self.internal_error_msg(e)
-      "Internal error. Please report a bug report with logs.\n" \
-        "Details: #{e.message}\n" \
-        "Caller:  #{e.backtrace.first}"
+      msg = "Internal error. Please report a bug report with logs.\n"
+
+      if e.is_a?(ArgumentError) && e.message =~ /invalid byte sequence in UTF-8/
+        msg += "A string was encountered that is not valid in UTF-8.\n" \
```
jreidinger: does it make sense to have message prefixed by internal error problem? I think different start of message will be better if is it specific for this issue.

src/ruby/yast/wfm.rb
```diff
@@ -194,9 +194,17 @@ def self.ask_to_run_debugger?
...
+               "The system encoding is #{Encoding.locale_charmap.inspect}.\n" \
+               "Refer to https://www.suse.com/support/kb/doc?id=1234567.\n\n"
+        # FIXME: fill in TID number
+      end
+
+      msg + "Details: #{e.message}\n" \
+            "Caller:  #{e.backtrace.first}"
```
jreidinger: does it make sense this for Argument error problem? especially when we know message?

src/ruby/yast/wfm.rb
mvidner: In the end it still is an internal error and we still want a bug report with logs, so it is better to keep it.

src/ruby/yast/wfm.rb
mvidner: Yes, it is redundant. IMHO it is better to keep it than complicate the rescue handler even further.

jreidinger:
LGTM

mvidner:
(Jenkins build fails because OBS cannot see the IBS repo, then GitHub shows it as still pending. Weird)

PR #175, opened 2016-10-13
More friendly wfm call

jreidinger:


mvidner:
Thanks for resurrecting this!

But I don't see how this improves the situation when I use the wrong arguments. Can you show the error message before and after? Or even add a test, like in https://github.com/yast/yast-ruby-bindings/compare/SLE-12-SP1...mvidner:callfunction?expand=1 ?

PR #177, opened 2016-11-15
Add debugger hooks

lslezak:
Hook to the `UI::` calls and detect the debug magic key event, start the Ruby debugger when it is received.

#### Advantages

- The debugger can be started during installation or in installed system when needed, e.g. you see a strange value in a dialog, simply pressing `Shift`+`Ctrl`+`Alt`+`D` starts the debugger and you can inspect the internal state immediately

#### Disadvantages

- Only the Qt UI can send the debugging magic key (i.e. it does not work in ncurses)
- The debugger can be started only from `UI::UserInput()` and similar calls
  - If the code does not interact with user (e.g. it's waiting for an external command to finish or is cycling in an infinite loop) the debugger cannot be started

Although there are some disadvantages of this solution it is a big step forward for easier YaST debugging.

### DIFF
--- a/src/binary/Yast.cc
+++ b/src/binary/Yast.cc
@@ -169,12 +169,12 @@ void set_ruby_source_location(VALUE file, VALUE lineno)
 }
 
 /**
- * Is the function name an UI function which returns a symbol?
+ * Returns true if the function name in an UI user input function which returns
+ * a symbol.
  * @param  function_name name of the function
- * @return true if function_name is one of "UserInput", "TimeoutUserInput"
- *   or "PollInput"
+ * @return true/false
  */
-static bool ui_symbol_function(const char *function_name)
+static bool ui_input_function(const char *function_name)
 {
     return strcmp(function_name, "UserInput") == 0  ||
         strcmp(function_name, "TimeoutUserInput") == 0 ||
@@ -182,9 +182,9 @@ static bool ui_symbol_function(const char *function_name)
 }
 
 /**
- * Is the YCP value :debugHotkey symbol?
- * @param  val YCP value
- * @return true when val == :debugHotkey
+ * Returns true if the parameter is the :debugHotkey symbol.
+ * @param  val YCPSymbol returned from an UI input call
+ * @return true/false
  */
 static bool is_debug_symbol(YCPValue val)
 {
@@ -193,20 +193,19 @@ static bool is_debug_symbol(YCPValue val)
 }
 
 /**
- * Is the function name an UI function which returns a map?
+ * Returns true if the function_name is "WaitForEvent"
  * @param  function_name name of the function
- * @return true if function_name is "WaitForEvent"
+ * @return true/false
  */
-static bool ui_hash_function(const char *function_name)
+static bool ui_input_event_function(const char *function_name)
 {
     return strcmp(function_name, "WaitForEvent") == 0;
 }
 
 /**
- * Is the YCP value a map containg a debug UI event?
- * @param  val YCP value
- * @return true if val is map and map["EventType"] == "DebugEvent" &&
- *   map["ID"] == :debugHotkey
+ * Returns true if the input is a debug UI event.
+ * @param  val YCPMap returned from the UI::WaitForEvent call
+ * @return true/false
  */
 static bool is_debug_event(YCPValue val)
 {
@@ -234,7 +233,7 @@ static bool is_debug_event(YCPValue val)
  */
 static void start_ruby_debugger()
 {
-    y2warning("Starting the Ruby debugger...");
+    y2milestone("Starting the Ruby debugger...");
 
     rb_require("yast/debugger");
     // call "Yast::Debugger.start"
@@ -344,8 +343,8 @@ ycp_module_call_ycp_function(int argc, VALUE *argv, VALUE self)
     if (strcmp(namespace_name, "UI") == 0)
     {
         if (
-            (ui_symbol_function(function_name) && is_debug_symbol(res)) ||
-            (ui_hash_function(function_name) && is_debug_event(res))
+            (ui_input_function(function_name) && is_debug_symbol(res)) ||
+            (ui_input_event_function(function_name) && is_debug_event(res))
         )
         {
           y2milestone("UI::%s() caught magic debug key: %s", function_name, res->toString().c_str());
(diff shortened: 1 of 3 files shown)


src/binary/Yast.cc
```diff
@@ -166,6 +168,81 @@ void set_ruby_source_location(VALUE file, VALUE lineno)
...
+ */
+static void start_ruby_debugger()
+{
+    y2warning("Starting the Ruby debugger...");
+
+    rb_require("yast/debugger");
+    // call "Yast::Debugger.start"
+    VALUE module = rb_const_get(rb_cObject, rb_intern("Yast"));
+    VALUE klass = rb_const_get(module, rb_intern("Debugger"));
+    rb_funcall(klass, rb_intern("start"), 0);
```
jreidinger: can it crash? if so, that it need to wrapped by default crash handler otherwise it segfault.

src/binary/Yast.cc
```diff
@@ -166,6 +168,81 @@ void set_ruby_source_location(VALUE file, VALUE lineno)
   YaST::ee.setLinenumber(FIX2INT(lineno));
 }
 
+/**
+ * Is the function name an UI function which returns a symbol?
```
jreidinger: does it really check all methods that return symbol? I think it is only events that poll events and return it as symbol, not?

src/binary/Yast.cc
```diff
@@ -166,6 +168,81 @@ void set_ruby_source_location(VALUE file, VALUE lineno)
...
+        event_id->asSymbol()->symbol() == "debugHotkey";
+}
+
+/**
+ * Start the Ruby debugger, it calls "Yast::Debugger.start" Ruby code.
+ * See file ../ruby/yast/debugger.rb for more details.
+ */
+static void start_ruby_debugger()
+{
+    y2warning("Starting the Ruby debugger...");
```
jreidinger: why it is warning? what is going wrong?

src/binary/Yast.cc
```diff
@@ -166,6 +168,81 @@ void set_ruby_source_location(VALUE file, VALUE lineno)
...
+ * @return true when val == :debugHotkey
+ */
+static bool is_debug_symbol(YCPValue val)
+{
+    return !val.isNull() && val->isSymbol() &&
+        val->asSymbol()->symbol() == "debugHotkey";
+}
+
+/**
+ * Is the function name an UI function which returns a map?
```
jreidinger: same as above for above regarding documentation. And in fact I do not like also name of method. What about having something like `ui_event_function` and `ui_input_function`? it makes more sense for me.

src/binary/Yast.cc
```diff
@@ -166,6 +168,81 @@ void set_ruby_source_location(VALUE file, VALUE lineno)
...
+ * @return true if function_name is "WaitForEvent"
+ */
+static bool ui_hash_function(const char *function_name)
+{
+    return strcmp(function_name, "WaitForEvent") == 0;
+}
+
+/**
+ * Is the YCP value a map containg a debug UI event?
+ * @param  val YCP value
```
jreidinger: quite useless comment. In fact it expect result of UI event. So something like `@param  val YCPMap response from UI Event call` ? or at least UI event map

src/binary/Yast.cc
```diff
@@ -166,6 +168,81 @@ void set_ruby_source_location(VALUE file, VALUE lineno)
...
+static bool ui_symbol_function(const char *function_name)
+{
+    return strcmp(function_name, "UserInput") == 0  ||
+        strcmp(function_name, "TimeoutUserInput") == 0 ||
+        strcmp(function_name, "PollInput") == 0;
+}
+
+/**
+ * Is the YCP value :debugHotkey symbol?
+ * @param  val YCP value
```
jreidinger: same as below

src/binary/Yast.cc
```diff
@@ -166,6 +168,81 @@ void set_ruby_source_location(VALUE file, VALUE lineno)
...
+ */
+static bool ui_symbol_function(const char *function_name)
+{
+    return strcmp(function_name, "UserInput") == 0  ||
+        strcmp(function_name, "TimeoutUserInput") == 0 ||
+        strcmp(function_name, "PollInput") == 0;
+}
+
+/**
+ * Is the YCP value :debugHotkey symbol?
```
jreidinger: hmm, I prefer here behavior, so something like `returns true if input is debug hotkey`
It is more flexible if we decide to rename it or having more symbols. Then contract of this function is to detect it and it is bug if it is not updated. Otherwise it looks like it work as expected, but in fact do not work.

lslezak: Um, what should be then the `@return` description? `returns true if input is debug hotkey` would be a duplication...

src/binary/Yast.cc
lslezak: Um, it should not crash, the `Yast::Debugger.start` is quite robust (handles missing `byebug` gem internally, etc...). If it by chance crashes then YaST is probably already so broken then a segfault is inevitable anyway...

Invoking the debugger is not an usual action, it's a kind of a corner case, so I'd keep it simple. We can always improve it later when we find a real case when it crashes...

src/binary/Yast.cc
```diff
@@ -166,6 +168,80 @@ void set_ruby_source_location(VALUE file, VALUE lineno)
   YaST::ee.setLinenumber(FIX2INT(lineno));
 }
 
+/**
+ * Returns true if the function name in an UI user input function which returns
```
jreidinger: typo `in` instead of `is`

src/binary/Yast.cc
```diff
@@ -166,6 +168,80 @@ void set_ruby_source_location(VALUE file, VALUE lineno)
...
+ */
+static bool ui_input_function(const char *function_name)
+{
+    return strcmp(function_name, "UserInput") == 0  ||
+        strcmp(function_name, "TimeoutUserInput") == 0 ||
+        strcmp(function_name, "PollInput") == 0;
+}
+
+/**
+ * Returns true if the parameter is the :debugHotkey symbol.
```
jreidinger: still I think it should be be exact symbol value in function name as I expect function provide some abstraction. maybe something like `Returns true if the input symbol starts debugging.`

jreidinger:
LGTM

PR #178, opened 2016-11-15
Debugger - backport to SP2

lslezak:
This is basically a backport of #177 to SP2.

Additional changes:

- Keep the same version, this is not an urgent bug fix. It will be released later with some other real fix.


jreidinger:
LGTM

PR #179, opened 2016-11-15
Debugger hooks

lslezak:
Hook to the `UI::` calls and detect the debug magic key event, start the Ruby debugger when it is received.

# Blog Text (Proposal)

We have improved the Ruby debugger integration in YaST. So far you could start the debugger using the `y2debugger=1` boot option or by setting the `Y2DEBUGGER=1` environment variable. The new feature allows starting the Ruby debugger also later when the YaST module is already running.

Simply press `Shift`+`Ctrl`+`Alt`+`D` magic key combination (`D` as debug) and it will start the Ruby debugger. It works during installation and also in installed system (just make sure the `byebug` Ruby gem is installed).

Unfortunately this new feature works only in the Qt UI, the ncurses UI is not supported (it currently does not handle the magic keyboard combination).

ruby_debugger


jreidinger:
LGTM

PR #180, opened 2016-12-15
Do not crash when empty.mo is missing

imobachgs:
Try to be more robust to avoid [bsc#1014458](https://bugzilla.suse.com/show_bug.cgi?id=1014458).

It looks like the problem is related to the update of fast_gettext. Once that package has been updated, the`/usr/lib64/ruby/gems/2.1.0/gems/fast_gettext-0.8.1/lib/fast_gettext/vendor/empty.mo` does not exist anymore (now is `/usr/lib64/ruby/gems/2.1.0/gems/fast_gettext-0.9.2/lib/fast_gettext/vendor/empty.mo`).

I think that the crash comes from [this line](https://github.com/grosser/fast_gettext/blob/fa8da8d6947108955450b1d0b515e32a36ee5c38/lib/fast_gettext/translation_repository/mo.rb#L41) which calls [MoFile.empty](https://github.com/grosser/fast_gettext/blob/fa8da8d6947108955450b1d0b515e32a36ee5c38/lib/fast_gettext/mo_file.rb#L39).

I haven't been able to reproduce it, but I think that this patch should solve the potential issue. I want to write some tests but I would like to have some early feedback.

### DIFF
--- a/src/ruby/yast/i18n.rb
+++ b/src/ruby/yast/i18n.rb
@@ -53,6 +53,12 @@ module Yast
         end
       end
       FastGettext::Translation::_ str
+    rescue Errno::ENOENT => error
+      if defined?(log) && log.respond_to?(:warn)
+        log.warn("File not found when translating '#{str}' on textdomain #{@my_textdomain}'. "\
+                 "Error: #{error}. Backtrace: #{error.backtrace}")
+      end
+      str
     end
 
     # No translation, only marks the text to be found by gettext when creating POT file,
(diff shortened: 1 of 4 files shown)


src/ruby/yast/i18n.rb
```diff
@@ -53,6 +53,12 @@ def _(str)
         end
       end
       FastGettext::Translation::_ str
+    rescue Errno::ENOENT => error
+      if defined?(log) && log.respond_to?(:warn)
```
jreidinger: you do not need to depend on log and call `Yast.y2warning` directly https://github.com/yast/yast-ruby-bindings/blob/master/src/ruby/yast/logger.rb#L58

just require logger. You do not need to include it.

jreidinger:
in general it looks like reasonable approach.

src/ruby/yast/i18n.rb
imobachgs: Oh, thanks for that tip!

imobachgs:
Thanks for the review!

I've refactored the code a little bit because there was bug. If we add a textdomain that does not exists and the `empty.mo` file is missing, we'll stop searching for a translation as soon as we get try that textdomain. Now, in such situation, the error is logged and the next textdomain is tried.

jreidinger:
LGTM

imobachgs:
Thanks!

PR #181, opened 2016-12-21
improve exception get from modules

jreidinger:
https://bugzilla.suse.com/show_bug.cgi?id=932331

depends on https://github.com/yast/yast-core/pull/115

example output from failed rspec when import failed with inner exception:

```
  1) Yast.import target module raises exception raises exception with details from original one
     Failure/Error: expect{ Yast.import "RaisingModule" }.to raise_error(/Testing exception/)

     RuntimeError:
       Failed to load Module 'RaisingModule' due to: Testing exception
     # ./test_module/modules/RaisingModule.rb:1:in `<top (required)>'
     # /home/jreidinger/yast/ruby-bindings/src/ruby/yast/yast.rb:168:in `import_pure'
     # /home/jreidinger/yast/ruby-bindings/src/ruby/yast/yast.rb:168:in `import'
     # ./yast_spec.rb:53:in `block (5 levels) in <top (required)>'
     # ./yast_spec.rb:53:in `block (4 levels) in <top (required)>'
```


mvidner:
Regarding the Travis failure: one cause is the dependence on new API in yast2-core, but there is also .travis.yml referencing `tests/ruby` which got removed, and some RuboCop nitpicks

jreidinger:
@mvidner updated. now rubocop green and test fails due to not build yastx, which is related to missing core

src/binary/Y2RubyComponent.cc
```diff
@@ -75,16 +76,30 @@ Y2Namespace *Y2RubyComponent::import (const char* name)
...
-  try {
-    YRuby::loadModule (args);
+  if (YRuby::loadModule (args))
+  {
     y2debug("Module '%s' loaded", name);
     // introspect, create data structures for the interpreter
     Y2Namespace * res = new YRubyNamespace (name);
     namespaces[name] = res;
     return res;
-  } catch (exception& e) {
```
mvidner: You remove a `catch` but remove no `throw`. This will fail in unexpected ways I think.

jreidinger: well, in fact it cannot throw, this catch is never called because:

1. ruby is pure C, so no exceptions
2. YNamespace do not throw

so I think it is just old bogus code and I do not expect anything to be raised here.

mvidner: 1 and 2 is correct, but YRubyNamespace does throw:
- https://github.com/yast/yast-ruby-bindings/blob/e41a0a1385c2ad6183ee6d903d59e5d148093205/src/binary/YRubyNamespace.cc#L312
- https://github.com/yast/yast-ruby-bindings/blob/e41a0a1385c2ad6183ee6d903d59e5d148093205/src/binary/YRubyNamespace.cc#L339

jreidinger: @mvidner ah, I overlook that construct_table call. So in such case I also suggest to create error namespace with message so it is clear what is problem and backtrace from C++ exception. Do you agree?

mvidner: Yes, that sounds good

PR #182, opened 2017-01-03
Do not crash when empty.mo is missing (SP1)

imobachgs:
Merge #180 into `SLE-12-SP1` branch.

### DIFF
--- a/tests/ruby/i18n_spec.rb
+++ b/tests/ruby/i18n_spec.rb
@@ -24,5 +24,78 @@ module Yast
         expect(Nn_(singular, plural, count)).to eq [singular, plural, count]
       end
     end
+
+    describe "._" do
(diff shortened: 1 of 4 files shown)


tests/ruby/i18n_spec.rb
```diff
@@ -24,5 +24,78 @@ module Yast
...
+      end
+
+      it "returns the translated string" do
+        allow(FastGettext).to receive(:key_exist?).and_return(true)
+        expect(FastGettext::Translation).to receive(:_).with(SINGULAR)
+          .and_return(TRANSLATED)
+        expect(_(SINGULAR)).to eq(TRANSLATED)
+      end
+
+      context "when FastGettext throws an exception" do
```
jreidinger: well, in fact it is not "throws an exception" but one specific type of exception.

PR #187, opened 2017-01-09
Drop add api

jreidinger:


jreidinger:
thanks for English improving

Issue #82, opened 2013-09-27
SCR builtins should accept strings in place of paths too


mvidner:
We should have noticed it in YCP Killer, but oh well. I have realized that the code always uses the `path` helper as the 1st argument to SCR builtins:

``` ruby
  SCR.Write(path(".etc.sysctl_conf"), nil) 
```

They should accept plain strings and convert them to paths internally

``` ruby
  SCR.Write(".etc.sysctl_conf", nil) 
```

PR #188, opened 2017-01-09
Allow String where Yast::Path is needed in SCR and WFM methods

jreidinger:
fixes https://github.com/yast/yast-ruby-bindings/issues/82


src/ruby/yast/scr.rb
```diff
@@ -100,16 +108,24 @@ def self.UnregisterAllAgents
...
       # caller(1) is one of the functions above
       res = caller(2, 1).first.match(BACKTRACE_REGEXP)
       call_builtin(res[1], res[2].to_i, *args)
     end
+
+    private_class_method def self.ensure_path(path)
+      return Path.new(path) if path.is_a?(::String)
+      return path if path.is_a?(Yast::Path)
+
+      raise ArgumentError, "argument '#{path.inspect}' is not path or string"
```
mvidner: I think it is better without the quotes, `inspect` will do a good job already. Plus exact classes, plus grammar.
```ruby
raise "Argument #{path.inspect} is neither a ::String or a Yast::Path"
```

src/ruby/yast/wfm.rb
```diff
@@ -263,5 +266,13 @@ def self.run_client(client)
         return false
       end
     end
+
+    private_class_method def self.ensure_path(path)
+      return Path.new(path) if path.is_a?(::String)
+      return path if path.is_a?(Yast::Path)
+
+      raise ArgumentError, "argument '#{path.inspect}' is not path or string"
```
mvidner: Uh, again? Repeating just twice is OK...

but here's a suggestion to save this repetition: move replace `ensure_path` to `Path`, perhaps even to `Path.new` ?
https://github.com/yast/yast-ruby-bindings/blob/87cb0136852c976ac72659504f294ffb07d45aaa/src/ruby/yast/path.rb#L10

```rb
    def initialize(value)
      return value.clone if value.is_a?(Yast::Path)
      if !path.is_a?(::String)
        raise ArgumentError, "Argument #{value.inspect} is neither a ::String or a Yast::Path"
      end

      @components = []
      load_components value
    end
```

AFAIK Yast::Path is not checking that it gets a String, and adding that could be useful(?)

PR #191, opened 2017-02-22
Y2base replacement

jreidinger:
Part of [Hack Week 15: More ruby in YaST](https://hackweek.suse.com/15/projects/more-ruby-in-yast)

related pull requests:

- https://github.com/yast/yast-installation/pull/529
- https://github.com/yast/yast-yast2/pull/539
- https://github.com/yast/yast-core/pull/119

### DIFF
--- a/package/yast2-ruby-bindings.changes
+++ b/package/yast2-ruby-bindings.changes
@@ -1,4 +1,21 @@
 -------------------------------------------------------------------
+Tue Feb 28 14:21:06 UTC 2017 - jreidinger@suse.com
+
+- Add new y2base as ruby script which have several advantages
+  (bsc#1027181):
+-- ruby is no longer bundled to YaST2, so it will make adopting
+   new ruby version much smoother
+-- show popup when signal is received and also allow YaST modules
+   to handle more gracefully when signal is raised
+-- better argument passing to y2base itself e.g.
+   y2base foo -S '(bar)' qt → y2base foo --arg bar qt
+-- ruby infrastructure can be used directly on y2base like debugger
+   or profiler
+-- majority of y2base functionality is covered by unit tests
+   allowing easier and less error-prone future changes
+- 3.2.6
+
+-------------------------------------------------------------------
 Mon Jan  9 14:18:30 UTC 2017 - jreidinger@suse.com
 
 - allow in Yast::SCR and Yast::WFM to have string as first
--- a/package/yast2-ruby-bindings.changes
+++ b/package/yast2-ruby-bindings.changes
@@ -1,18 +1,18 @@
 -------------------------------------------------------------------
 Wed Mar 22 15:45:50 UTC 2017 - jreidinger@suse.com
 
-- Add new y2base as ruby script which have several advantages
+- Add new y2start as ruby script which have several advantages
   (bsc#1027181):
 -- Ruby is no longer embedded into YaST (instead we only use binary
    extensions), so there is less chance of things breaking with
    a new Ruby version
 -- show popup when signal is received and also allow YaST modules
    to handle more gracefully when signal is raised
--- better argument passing to y2base itself e.g.
-   y2base foo -S '(bar)' qt → y2base foo --arg bar qt
--- ruby infrastructure can be used directly on y2base like debugger
+-- better argument passing to y2start itself e.g.
+   y2base foo -S '(bar)' qt → y2start foo --arg bar qt
+-- ruby infrastructure can be used directly on y2start like debugger
    or profiler
--- majority of y2base functionality is covered by unit tests
+-- majority of y2start functionality is covered by unit tests
    allowing easier and less error-prone future changes
 - 3.2.10
 
@@ -46,7 +46,6 @@ Thu Mar  9 15:57:00 UTC 2017 - ancor@suse.com
 - Display standard output and error channels in case of failure
   running the Ncurses integration tests in a headless system.
   Needed to debug the error produced by the fix to bsc#1021743
->>>>>>> origin/master
 - 3.2.6
 
 -------------------------------------------------------------------
--- a/package/yast2-ruby-bindings.changes
+++ b/package/yast2-ruby-bindings.changes
@@ -1,5 +1,5 @@
 -------------------------------------------------------------------
-Tue Feb 28 14:21:06 UTC 2017 - jreidinger@suse.com
+Wed Mar 22 15:45:50 UTC 2017 - jreidinger@suse.com
 
 - Add new y2base as ruby script which have several advantages
   (bsc#1027181):
@@ -14,6 +14,39 @@ Tue Feb 28 14:21:06 UTC 2017 - jreidinger@suse.com
    or profiler
 -- majority of y2base functionality is covered by unit tests
    allowing easier and less error-prone future changes
+- 3.2.10
+
+-------------------------------------------------------------------
+Wed Mar 22 14:45:50 UTC 2017 - jreidinger@suse.com
+
+- Use more friendly exception when an invalid Yast::Path is
+  constructed (one with a component starting or ending with
+  a dash; bsc#1028081)
+- 3.2.9
+
+-------------------------------------------------------------------
+Wed Mar 22 14:41:50 UTC 2017 - jreidinger@suse.com
+
+- add method Yast::WFM.scr_root to get scr changed root directory
+  (needed for testing fix for bsc#1023204)
+- 3.2.8
+
+-------------------------------------------------------------------
+Tue Mar 14 16:25:59 UTC 2017 - ancor@suse.com
+
+- Use tmux instead of screen to run the Ncurses integration test
+  in headless systems (skip the test if tmux is not available).
+  The version of screen introduced to fix bsc#1021743 was causing
+  the test to fail.
+- 3.2.7
+
+-------------------------------------------------------------------
+Thu Mar  9 15:57:00 UTC 2017 - ancor@suse.com
+
+- Display standard output and error channels in case of failure
+  running the Ncurses integration tests in a headless system.
+  Needed to debug the error produced by the fix to bsc#1021743
+>>>>>>> origin/master
 - 3.2.6
 
 -------------------------------------------------------------------
@@ -40,7 +73,6 @@ Wed Jan  4 12:36:06 UTC 2017 - jreidinger@suse.com
 - Throw exception with more detailed error specification when
   loading namespace to component system failed (bsc#932331)
 - 3.2.4
-
 
 -------------------------------------------------------------------
 Thu Dec 15 16:29:13 UTC 2016 - igonzalezsosa@suse.com
--- a/package/yast2-ruby-bindings.changes
+++ b/package/yast2-ruby-bindings.changes
@@ -3,8 +3,9 @@ Tue Feb 28 14:21:06 UTC 2017 - jreidinger@suse.com
 
 - Add new y2base as ruby script which have several advantages
   (bsc#1027181):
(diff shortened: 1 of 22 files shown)


jreidinger:
changelog entry proposal:

"- Start YaST2 from ruby instead of embedding it (bsc#1027181)"
"- add popup when signal kills YaST"
"- when yast crashes suggest to report bug with logs obtained with save_y2logs script"

src/ruby/yast/wfm.rb
```diff
@@ -213,6 +213,61 @@ def self.call_builtin_wrapper(*args)
             "Caller:  #{e.backtrace.first}"
     end
 
+    private_class_method def self.handle_signal_exception(e)
+      signame = Signal.signame(e.signo)
+      msg = "YaST receive signal %s and will exit.\n If killing of YaST is " \
+        "not requested then please report a bug report with logs obtained from " \
+        "save_y2logs script." % signame
```
mvidner: - How does this message look in a 80x25 terminal?
- "YaST receive**d** **a** %s signal and will exit"
- What do you mean "if killing of YaST is not requested"? I think it depends on the signal, and this is the part that could annoy users the most. We have to get this detail right.

src/ruby/yast/y2base.rb
```diff
@@ -0,0 +1,131 @@
+module Yast
+  module Y2Base
```
mvidner: IMHO having both a program and a module named y2base is confusing. I would accept it if the program was a one-liner, or we should rename the module.

src/ruby/yast/y2base.rb
```diff
@@ -0,0 +1,131 @@
...
+
+    private_class_method def self.parse_client_options(args)
+      res = {}
+      res[:params] = []
+      loop do
+        return res unless option?(args.first)
+
+        arg = args.shift
+        case arg
+        when "-a", "--arg"
```
mvidner: changelog:
- changed the syntax for client arguments
y2base foo -S bar qt → y2base foo --arg bar qt

src/ruby/yast/y2base.rb
```diff
@@ -0,0 +1,131 @@
...
+
+    private_class_method def self.signal_handler(name)
+      Signal.trap(name, "IGNORE")
+
+      $stderr.puts "YaST got signal #{name}."
+
+      signal_log_open do |f|
+        f.puts "=== #{Time.now} ==="
+        f.puts "YaST got signal #{name}."
+        # TODO: print stored debug logs
```
mvidner: So this breaks that feature

src/ruby/yast/y2base.rb
```diff
@@ -0,0 +1,131 @@
...
+      "Examples:\n" \
+      "y2base installation qt\n" \
+      "    Start binary y2base with intallation.ycp as client and qt as server\n" \
+      "y2base installation -a initial qt\n" \
+      "    Provide parameter initial for client installation\n" \
+      "y2base installation qt -geometry 800x600\n" \
+      "    Provide geometry information as specific server options\n"
+    end
+
+    # so how works signals now in ruby?
```
mvidner: raise GrammarError :-/ please fix that yourself

package/yast2-ruby-bindings.spec
```diff
@@ -99,6 +99,7 @@ cd -
 
 %files
 %defattr (-, root, root)
+%{yast_ybindir}/y2base
```
jreidinger: yes, I plan to add it when approved, so I use correct yast2-core version to conflict with.

src/binary/Yast.cc
```diff
@@ -521,6 +521,10 @@ static VALUE ui_set_component(VALUE self, VALUE name)
 
 static void init_ui()
 {
+  // init_ui is needed only if run via rspec, if y2base is used ENV is initialized
```
jreidinger: in fact, here y2base stands for both :)
what comment want to say, is that code is used only for test purpose. Real code initialize it on its own, but as init_ui is called early, we need to skip it for production code.

src/ruby/yast/wfm.rb
```diff
@@ -201,7 +201,7 @@ def self.call_builtin_wrapper(*args)
     # @param [Exception] e the caught exception
     # @return [String] human readable exception description
     private_class_method def self.internal_error_msg(e)
-      msg = "Internal error. Please report a bug report with logs.\n"
+      msg = "Internal error. Please report a bug report with logs from save_y2logs script.\n"
```
jreidinger: what about adding "Run save_y2logs to get complete logs." ? otherwise it can get idea that this script do such reporting, which is not true.

src/ruby/yast/wfm.rb
jreidinger: will check it in small terminal. Type I fix....Well part with killing probably affect only SIGTERM, but e.g. even systemd can send such signal.

And one more idea we discussed early, that maybe we should use timed popup, to kill it in reasonable time after getting sigterm

mvidner:
What will we write about this to the team blog? Why did we do this?

I think we need to sell this, because the signal change is big enough that it will cause some problems, plus we will surely overlook some other bugs.

src/ruby/yast/wfm.rb
```diff
@@ -213,6 +213,61 @@ def self.call_builtin_wrapper(*args)
...
+        Yast.import "Report"
+        Report.Error(msg)
+      end
+    rescue Exception => e
+      Builtins.y2internal("Error reporting failed with '%1' and backtrace %2",
+        e.message,
+        e.backtrace)
+    end
+
+    private_class_method def self.check_type!(result)
```
jreidinger: fine for me.

src/ruby/yast/wfm.rb
```diff
@@ -221,47 +276,17 @@ def self.run_client(client)
...
-        allowed_types.delete(::Object) # remove generic type for any
-
-        # check if response is allowed
-        allowed = allowed_types.any? { |t| result.is_a? t }
-
-        raise "Invalid type #{result.class} from client #{client}" unless allowed
+        check_type!(result)
 
         return result
+      rescue SystemExit =>
```
jreidinger: make sense

src/ruby/yast/wfm.rb
```diff
@@ -221,47 +276,17 @@ def self.run_client(client)
...
+        raise # something call system exit so do not block it
+      rescue SignalException => e
+        handle_signal_exception(e)
+        exit(1)
```
jreidinger: so 16 make sense here?

src/ruby/yast/y2base.rb
jreidinger: well, goal is to have module that can be tested...maybe `Y2BaseHelpers` is good enough name?

src/ruby/yast/y2base.rb
jreidinger: yes. currently there is no way how to call blanik from ruby bindings.

src/ruby/yast/y2base.rb
jreidinger: in fact it was `y2base foo -S '(bar)' qt → y2base foo --arg bar qt`

jreidinger:
@mvidner regarding benefits:

1) no more bundling of ruby interpreter, so easier future upgrade to new ruby versions
2) popup shown when signal received and not silent exit of qt dialog. ( for ncurses it prints something to stderr )

jreidinger:
and 3. better argument passing to y2base :)

package/yast2-ruby-bindings.spec
mvidner: No, this way we forget it. Add it right away with the currently correct version.

jreidinger: done

package/yast2-ruby-bindings.spec
```diff
@@ -39,8 +39,10 @@ Requires:       rubygem(%{rb_default_ruby_abi}:fast_gettext)
 %endif
 BuildRequires:  ruby-devel
 # new ErrorNamespace needed
-Requires:       yast2-core >= 3.2.1
-BuildRequires:  yast2-core-devel >= 3.2.1
+Requires:       yast2-core >= 3.2.3
+BuildRequires:  yast2-core-devel >= 3.2.3
+# replacement of y2base with its ruby implementation
+Conflicts:      yast2-core <= 3.2.2
```
mvidner: I completely missed the fact that we already require a recent version of the package. That implies conflicting with lower versions, so Conflicts can be removed.

package/yast2-ruby-bindings.changes
```diff
@@ -1,4 +1,21 @@
 -------------------------------------------------------------------
+Tue Feb 28 14:21:06 UTC 2017 - jreidinger@suse.com
+
+- Add new y2base as ruby script which have several advantages
+  (bsc#1027181):
+-- ruby is no longer bundled to YaST2, so it will make adopting
+   new ruby version much smoother
```
mvidner: > Ruby is no longer embedded into YaST (instead we only use binary extensions), so there is less chance of things breaking with a new Ruby version

src/ruby/yast/wfm.rb
```diff
@@ -213,6 +214,68 @@ def self.call_builtin_wrapper(*args)
             "Caller:  #{e.backtrace.first}"
     end
 
+    # Handles a SignalExpection
+    private_class_method def self.handle_signal_exception(e)
+      signame = Signal.signame(e.signo)
+      msg = "YaST received a signal %s and will exit.\n" % signame
+      # sigterm are often sent by user
+      if e.signo == 15
+        msg += "If termination is not send by user then please report a bug report with logs.\n"
```
mvidner: sent

src/ruby/yast/y2base_helpers.rb
```diff
@@ -0,0 +1,132 @@
...
+      ret[:client_name] = args.shift or raise "Missing client name."
+      ret[:client_options] = parse_client_options(args)
+      ret[:server_name] = args.shift or raise "Missing server name."
+      ret[:server_options] = args
+
+      ret
+    end
+
+    def self.help
+      "Usage: y2base [GenericOpts] Client [ClientOpts] Server " \
```
mvidner: src/ruby/yast/y2base_helpers.rb:25: warning: unused literal ignored

spot the bug

mvidner: I have fixed it.

package/yast2-ruby-bindings.changes
```diff
@@ -1,4 +1,22 @@
...
+
+- Add new y2base as ruby script which have several advantages
+  (bsc#1027181):
+-- Ruby is no longer embedded into YaST (instead we only use binary
+   extensions), so there is less chance of things breaking with
+   a new Ruby version
+-- show popup when signal is received and also allow YaST modules
+   to handle more gracefully when signal is raised
+-- better argument passing to y2base itself e.g.
+   y2base foo -S '(bar)' qt → y2base foo --arg bar qt
```
mvidner: Replace y2base with y2start also in the changes entry almost everywhere, except when quoting old usage:

... to y2start itself, e.g.
y2base foo -S '(bar)' '(baz)' qt → y2start foo --arg bar --arg baz qt

package/yast2-ruby-bindings.changes
```diff
@@ -28,6 +46,7 @@ Thu Mar  9 15:57:00 UTC 2017 - ancor@suse.com
 - Display standard output and error channels in case of failure
   running the Ncurses integration tests in a headless system.
   Needed to debug the error produced by the fix to bsc#1021743
+>>>>>>> origin/master
```
mvidner: Oops

PR #192, opened 2017-02-23
add method to get scr root

jreidinger:


package/yast2-ruby-bindings.changes
```diff
@@ -1,4 +1,11 @@
 -------------------------------------------------------------------
+Wed Mar 22 14:41:50 UTC 2017 - jreidinger@suse.com
+
+- add method Yast::WFM.scr_root get scr changed root directory
```
imobachgs: *to get*

src/ruby/yast/wfm.rb
```diff
@@ -119,6 +119,19 @@ def self.scr_chrooted?
       SCRGetName(SCRGetDefault()) != "scr"
     end
 
+    # Returns root on which scr operated e.g. "/" when scr not switched
```
imobachgs: operates

imobachgs: I guess it's a string, but adding a `@return` description would be fine.

PR #193, opened 2017-03-07
fix crash when invalid path is passed

jreidinger:
fix for L3 https://bugzilla.suse.com/show_bug.cgi?id=1028081

customer reviewing it now


src/ruby/yast/path.rb
```diff
@@ -119,7 +119,7 @@ def load_components(value)
 
     def invalid_buffer?(buffer)
       if buffer.start_with?("-") || buffer.end_with?("-")
-        Yast.y2error "Cannot have dash before or after dot '#{value}'"
+        Yast.y2error "Cannot have dash before or after dot '#{buffer}'"
```
mvidner: Note to self: add this to my collection of bugs that are discovered by static analysis.

jreidinger: you mean "can be" discovered? and to be honest you cannot be ever sure, as method `value` can be dynamically defined ;)

package/yast2-ruby-bindings.changes
```diff
@@ -1,4 +1,9 @@
 -------------------------------------------------------------------
+Wed Mar  8 12:55:38 UTC 2017 - jreidinger@suse.com
+
+- Do not crash when invalid path is passed (bsc#1028081)
```
mvidner: This is good, but I am curious about the exact circumstances so I've come up with a more specific description:

> Do not crash when an invalid Yast::Path is constructed (one with a component starting or ending with a dash; bsc#1028081)

mvidner: The RPM version is unchanged on purpose?

jreidinger: yes, I do not plan to release it now, only if another fix appear as it is minor and usually root of issue is elsewhere ( in given bug report problem is in network ).

src/ruby/yast/path.rb
mvidner: "could be discovered", yes

mvidner:
So, before it raised a `NameError`. What does it do now? I think it continues with an empty path. That actually seems worse. Shouldn't we either

1. raise `ArgumentError`, or
2. allow such paths?

jreidinger:
@mvidner this behavior keeps original one from ycp. It is original intention as it is used in ycp builtin to construct path from string.

mvidner:
No, it does not keep the YCP behavior: back then, it was a lexer error: https://github.com/yast/yast-core/blob/Code-11-SP3/libycp/testsuite/tests/values/PathErr1.err
Now we let the code run, with a nonsensical object.

jreidinger:
@mvidner no, correct one is https://github.com/yast/yast-core/blob/Code-11-SP3/libycp/testsuite/tests/values/PathTopath.ycp and as string is unknown, then ti cannot know in advance how string will look like.

jreidinger:
@mvidner and also in bug report it is generated path, not static one.

mvidner:
Well, [Routing.rb](https://github.com/yast/yast-network/blob/46cd5b80124a69eec2807fc580c66fb8dd323653/src/modules/Routing.rb#L540) is using `path("ifroute-#{device}")` which is wrong. The `path` helper method is for YCP converted code and should only see literal strings.

The real question is: will network code be fixed to deal with the bug report? If not, the situation will be even worse, because it will work on a `YCP::Path.new(".")` (or `""`?).

jreidinger:
@mvidner so what you suggest? if we modify it to raise exception we can get a lot of bug reports ( but not probably as this is reported after long time ) or to return nil as it was before?

jreidinger:
@mvidner btw `path()` is not enhanced to load string by default. so it act like `topath`

PR #195, opened 2017-03-14
Run the libyui integration test in Travis

ancorgs:


Dockerfile
```diff
@@ -1,2 +1,5 @@
 FROM yastdevel/cpp
 COPY . /usr/src/app
+
+# Install tmux to make sure the libyui+YaST integration tests are run
+RUN zypper in tmux
```
jreidinger: failing in travis. It have to be run non-interactive. Otherwise it looks fine.

Dockerfile
```diff
@@ -1,2 +1,5 @@
 FROM yastdevel/cpp
 COPY . /usr/src/app
+
+# Install tmux to make sure the libyui+YaST integration tests are run
+RUN zypper --non-interactive in tmux
```
mvidner: I think this could be moved between FROM and COPY because tmux will change less often than our git and docker caching will slightly improve.

PR #198, opened 2017-05-02
Fix cli

jreidinger:
please see commit message in https://github.com/yast/yast-ruby-bindings/commit/c5c6b67a1b4225b4f7ffb8f135ef1dcee74e91ff for explanation


kobliha:
It seems that the change does not what it was expected to do, see https://bugzilla.suse.com/show_bug.cgi?id=1037891

PR #199, opened 2017-05-23
Title in installation

jreidinger:
tested manually in VM

### DIFF
--- a/src/y2start/y2start
+++ b/src/y2start/y2start
@@ -49,7 +49,9 @@ hostname = " @ #{hostname}" unless hostname.empty?
 
 title = "YaST2 - #{args[:client_name]}#{hostname}"
 # set title only if it is not CLI (bsc#1033993)
-Yast::UI.SetApplicationTitle(title) if args[:client_options][:params].empty?
+# modules do not have CLI arguments, except installation (bsc#1037891)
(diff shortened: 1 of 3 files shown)


src/y2start/y2start
```diff
@@ -49,7 +49,9 @@ hostname = " @ #{hostname}" unless hostname.empty?
 
 title = "YaST2 - #{args[:client_name]}#{hostname}"
 # set title only if it is not CLI (bsc#1033993)
-Yast::UI.SetApplicationTitle(title) if args[:client_options][:params].empty?
+# modules do not have CLI arguments, except installation (bsc#1037891)
+set_title = args[:client_options][:params].empty? || args[:client_name] == "installation"
```
imobachgs: So I guess that `args[:client_options]` and `args[:client_options][:params]` could not be `nil`, right?

jreidinger: yes, it is result of params parsing. Always defined.

jreidinger: see https://github.com/yast/yast-ruby-bindings/blob/master/src/ruby/yast/y2start_helpers.rb#L12

imobachgs: Just to be sure, thanks!

PR #201, opened 2017-07-04
Fix syslog

jreidinger:


src/y2start/y2start
```diff
@@ -39,6 +39,7 @@ if !Yast::WFM.ClientExists(args[:client_name])
 end
 
 
+NO_CLI_CLIENTS = ["installation", "view_anymsg"]
```
imobachgs: what about `freeze`?

jreidinger: possible..that reminds me that rubocop is not set to this file as it does not have .rb suffix

PR #203, opened 2017-07-12
Ruby24

jreidinger:


package/yast2-ruby-bindings.spec
```diff
@@ -17,7 +17,7 @@
 
 
 Name:           yast2-ruby-bindings
-Version:        3.2.14
+Version:        3.2.15
```
ancorgs: Inconsistent version jump with the changelog.

jreidinger: my fault, will fix it

ancorgs:
LGTM

PR #204, opened 2017-07-12
Ruby24

jreidinger:


mvidner:
@ancorgs well, if Yast::Integer makes you *really* sad then you have an opportunity to et better, because `Yast.import "Integer"` has only 12 occurrences in the codebase.

PR #205, opened 2017-08-04
Caasp merge

jreidinger:


jreidinger:
I suggest to simple delete branch. no interesting code changes here.

package/yast2-ruby-bindings.changes
```diff
@@ -3,14 +3,15 @@ Thu Dec 15 16:29:13 UTC 2016 - igonzalezsosa@suse.com
 
 - Do not crash when FastGettext is unable to find the empty.mo
   file (bsc#1014458)
-- 3.1.52
+- 3.1.51.2
```
teclator: It should not change

package/yast2-ruby-bindings.spec
```diff
@@ -17,7 +17,7 @@
 
 
 Name:           yast2-ruby-bindings
-Version:        3.1.52
+Version:        3.1.51.2
```
teclator: Why?

teclator: Already discussed in other PR, basically we use the caasp one or the minor one because if not it will not be available.

teclator: MMM what will happen if we do a fix for SLE-12-SP2? and we unified to 3.1.52 instead of 3.1.53?

jreidinger:
ok, will delete branch.

PR #207, opened 2017-11-03
Better eval backtrace

mvidner:
[bsc#1066290](https://bugzilla.suse.com/show_bug.cgi?id=1066290)


src/ruby/yast/convert.rb
```diff
@@ -37,7 +37,7 @@ module Convert
     # @!method  self.to_locale(        object )
     #   @return        [String, nil]  *object*, or `nil` if it is not a String
     Ops::SHORTCUT_TYPES.each do |type|
-      eval <<END
+      eval <<END, binding, __FILE__, __LINE__ + 1
```
jreidinger: well, with my current ruby knowledge, I propose to remove eval and instead use `define_method`. How do you see it? https://apidock.com/ruby/Module/define_method

src/ruby/yast/convert.rb
```diff
@@ -37,11 +37,9 @@ module Convert
     # @!method  self.to_locale(        object )
     #   @return        [String, nil]  *object*, or `nil` if it is not a String
     Ops::SHORTCUT_TYPES.each do |type|
-      eval <<END, binding, __FILE__, __LINE__ + 1
-        def self.to_#{type}(object)
-          convert object, :from => "any", :to => "#{type}"
-        end
-END
+      define_singleton_method("to_#{type}") do |object|
+        convert object, :from => "any", :to => "#{type}"
```
mvidner: While we're at this, can we simplify `"#{type}"` to `type`, which is a plain String?

src/ruby/yast/ops.rb
```diff
@@ -66,11 +66,9 @@ module Ops
     # @!method                     self.get_locale(       obj, idx, def )
     #   @return [String, nil]   {Convert.to_locale}({get}(obj, idx, def))
     Ops::SHORTCUT_TYPES.each do |type|
-      eval <<END, binding, __FILE__, __LINE__ + 1
-        def self.get_#{type}(object, indexes, default=nil, &block)
-          Yast::Convert.to_#{type} get(object, indexes, default, 1, &block)
-        end
-END
+      define_singleton_method("get_#{type}") do |object, indexes, default=nil, &block|
```
mvidner: Does a default value for a block argument work?? This (old) post says it does not: https://www.ruby-forum.com/topic/121058 Do we have a test for this case?

jreidinger: in new ruby it should, but I will add test.

PR #208, opened 2017-11-23
Hardcode the Ruby version in the y2start script (bsc#1068863)

lslezak:
- Hardcode the Ruby version in the y2start script to always use the Ruby version used at the build time
- Fixes [bsc#1068863](https://bugzilla.suse.com/show_bug.cgi?id=1068863)
- Updated the search for the Ruby versions (though it does not seem to be needed anymore)
- See the [Travis output](https://travis-ci.org/yast/yast-ruby-bindings/builds/306325291#L706)
- The initially trivial `sed` become quite complicated because of the cmake(shell(sed))) command escaping :worried: 
- 4.0.1


cmake/modules/FindRuby.cmake
```diff
@@ -15,7 +15,7 @@ if(RUBY_LIBRARY AND RUBY_INCLUDE_PATH)
    set(RUBY_FIND_QUIETLY TRUE)
 endif (RUBY_LIBRARY AND RUBY_INCLUDE_PATH)
 
-FIND_PROGRAM(RUBY_EXECUTABLE NAMES ruby ruby1.9 ruby19 ruby1.9.1 ruby2.1)
+FIND_PROGRAM(RUBY_EXECUTABLE NAMES ruby ruby2.4 ruby.ruby2.4 ruy2.5 ruby.ruby2.5)
```
aschnell: "ruy" and "ruby.ruby"?

jreidinger: this should be needed. We should search for only ruby. Version specific ones is problem as have version specific RbConfig which we do not want. We want only system one.

cmake/modules/FindRuby.cmake
```diff
@@ -54,7 +58,7 @@ FIND_PATH(RUBY_INCLUDE_PATH
 )
 
 FIND_LIBRARY(RUBY_LIBRARY
-  NAMES ruby ruby1.9 ruby-1.9.1 ruby2.1 ruby-2.1
+  NAMES ruby ruby2.4 ruby2.5
```
jreidinger: same here. We want only system one.

cmake/modules/FindRuby.cmake
lslezak: Oh, "ruy" is a typo. Thanks! But "ruby.ruby*" is actually correct.

PR #209, opened 2017-12-05
Title firstboot sp3

jreidinger:


src/y2start/y2start
```diff
@@ -38,7 +38,7 @@ if !Yast::WFM.ClientExists(args[:client_name])
   exit 1
 end
 
-NO_CLI_CLIENTS = ["installation", "view_anymsg"].freeze
+NO_CLI_CLIENTS = ["installation", "view_anymsg", "firstboot"].freeze
```
mvidner: The constant name is confusing. This should be named "modules that have GUI even though they have CLI arguments". How about GUI_BUT_HAVE_ARGS ?

PR #213, opened 2018-02-28
Workaround for y2makepot

lslezak:


src/ruby/yast/i18n.rb
```diff
@@ -1,6 +1,11 @@
 require "fast_gettext"
 require "logger"
 
+# just to make the y2makepot script happy
+'
+textdomain "example"
+'
+
```
jreidinger: well, ok, but please place it to bottom of file and not top, as it is just workaround.

PR #214, opened 2018-03-27
Log a warning if no text domain is configured (bsc#1081466)

lslezak:
- This helps with debugging so the missing `textdomain` calls can be easier found.
- Writing the caller to the log helps to find the place where the translation call was used incorrectly. Grepping the sources for the label is a bit tricky if it contains the `&` shortcut marker or `%s` placeholders.
- 4.0.5

### Example Log

I have tested this when fixing bug https://github.com/yast/yast-network/pull/619, now the problematic place is logged into y2log:

```
2018-03-27 13:31:49 <2> dhcp25(24521) [Ruby] yast/i18n.rb:47 No textdomain configured, ca
nnot translate text "Enable access using a &web browser"
2018-03-27 13:31:49 <2> dhcp25(24521) [Ruby] yast/i18n.rb:48 Called from: /usr/share/YaST
2/lib/y2remote/widgets/remote.rb:165:in `label'
```

### DIFF
--- a/package/yast2-ruby-bindings.changes
+++ b/package/yast2-ruby-bindings.changes
@@ -1,4 +1,11 @@
 -------------------------------------------------------------------
+Tue Mar 27 12:52:27 UTC 2018 - lslezak@suse.cz
+
+- Log a warning if no text domain is configured for translations,
+  this helps with debugging (improved as a part of bsc#1081466)
+- 4.0.5
+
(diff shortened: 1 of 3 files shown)


mvidner:
@lslezak Thank you! This is a classic from the How-primitive!-But-how-effective! department.

PR #216, opened 2018-05-28
log method name

jreidinger:
how it looks before change:
```
2018-05-28 13:10:29 <1> linux-vvcf.privatesite(12728) [Ruby] y2storage/dump_manager.rb:230 Devicegraph dump directory: /var/log/YaST2/storage
2018-05-28 13:10:29 <1> linux-vvcf.privatesite(12728) [Ruby] y2storage/dump_manager.rb:264 Dumping Devicegraph to 01-probed
```

and after:
```
2018-05-28 13:25:35 <1> linux-vvcf.privatesite(20263) [Ruby] y2storage/dump_manager.rb(lazy_init):230 Devicegraph dump directory: /var/log/YaST2/storage
2018-05-28 13:25:35 <1> linux-vvcf.privatesite(20263) [Ruby] y2storage/dump_manager.rb(dump_internal):264 Dumping Devicegraph to 01-probed
```

I tested if it breaks old testsuite and result is that it is broken. Example failed test:
```
[   10s] --- tests/Map.out	2018-05-11 06:32:41.000000000 +0000
[   10s] +++ tmp.out.Map	2018-05-28 11:35:08.562451210 +0000
[   10s] @@ -26,32 +26,32 @@
[   10s]  Return	$["A":1, "B":1]
[   10s]  Dump	Map::CheckKeys
[   10s]  Return	false
[   10s] -Log	Missing key: 1
[   10s] +Log	in CheckKeys):97 Missing key: 1
[   10s]  Return	false
[   10s] -Log	Missing key: nil
[   10s] +Log	in CheckKeys):97 Missing key: nil
[   10s]  Return	false
[   10s]  Return	true
[   10s] -Log	Missing key: 2
[   10s] +Log	in CheckKeys):97 Missing key: 2
[   10s]  Return	false
[   10s] -Log	Missing key: 1
[   10s] +Log	in CheckKeys):97 Missing key: 1
[   10s]  Return	false
[   10s]  Return	true
[   10s]  Return	true
[   10s] -Log	Missing key: `y
[   10s] +Log	in CheckKeys):97 Missing key: `y
[   10s]  Return	false
[   10s] -Log	Missing key: 1
[   10s] +Log	in CheckKeys):97 Missing key: 1
[   10s]  Return	false
[   10s]  Return	true
[   10s]  Return	true
[   10s]  Return	true
[   10s] -Log	Missing key: 2
[   10s] +Log	in CheckKeys):97 Missing key: 2
[   10s]  Return	false
[   10s] -Log	Missing key: 4
[   10s] -Log	Missing key: 5
[   10s] +Log	in CheckKeys):97 Missing key: 4
[   10s] +Log	in CheckKeys):97 Missing key: 5
[   10s]  Return	false
[   10s] -Log	Missing key: 2
[   10s] +Log	in CheckKeys):97 Missing key: 2
[   10s]  Return	false
[   10s] -Log	Missing key: `x
[   10s] +Log	in CheckKeys):97 Missing key: `x
```


mvidner:
FYI, I did a quick count of the old test cases (`tests/*.out` files) and grouped them by repositories.
Total:  945

```
github-checkout$ find -path \*/tests/\*.out | sort | cut -f3 -d/ | uniq -c | sort -n
      1 yast-audit-laf
      1 yast-fcoe-client
      1 yast-iplb
      1 yast-nis-client
      1 yast-printer
      2 yast-hardware-detection
      2 yast-multipath
      2 yast-slp-server
      3 yast-nis-server
      4 yast-mail
      4 yast-nfs-server
      4 yast-squid
      6 yast-http-server
      6 yast-sound
      7 yast-perl-bindings
      8 yast-samba-client
      9 yast-dns-server
      9 yast-nfs-client
      9 yast-sysconfig
     11 yast-dhcp-server
     19 yast-country
     35 yast-users
     37 yast-samba-server
     37 yast-storage
     71 yast-yast2
    655 yast-core
```

jreidinger:
@mvidner are you sure you have the latest checkout? as I am pretty sure old storage is dead and new one does not have anything in old testsuite.

jreidinger:
@mvidner and second note. We are not interested in yast-core as it does not use ruby and its bindings.

so it should be 945 - 655 - 37, so 253 if I count correctly.

lslezak:
I think `yast-users` and maybe some more can be ignored as they are written in Perl (i.e. not using Ruby bindings).

shundhammer:
The old testsuite must have some code where it replaces the y2log header of each line with a regexp; otherwise we'd get a zillion mismatches with the leading timestamp and PID. So it should be possible to extend that regexp to remove the function name as well. We have to be careful to use non-greedy matching, though, to avoid removing too much.

jreidinger:
in shared part of old testsuite it is done by this magic: https://github.com/yast/yast-testsuite/blob/master/skel/runtest.sh#L69

jreidinger:
It is now ready for review as we drop all old testsuite and experimental build of YaST:Head with change builds.

lslezak (approved):
OK, if all old testsuite is dropped.

Please announce the change in yast-devel@, some third party YaST modules (e.g. OES) still might use some old tests.

PR #219, opened 2018-10-23
Fix encoding sp4

jreidinger:


package/yast2-ruby-bindings.changes
```diff
@@ -1,3 +1,10 @@
+-------------------------------------------------------------------
+Tue Oct 23 09:52:53 UTC 2018 - jreidinger@suse.com
+
+- Fix crash caused by reading external files in ASCII encoding
+  (bsc#1111367)
+- 3.3.0
```
lslezak: Version conflict, 3.3.0 has been already released in SLE15 with a different fix: https://github.com/yast/yast-ruby-bindings/blob/SLE-15-GA/package/yast2-ruby-bindings.changes#L61

src/y2start/y2start
```diff
@@ -1,4 +1,4 @@
-#! /usr/bin/ruby
+#! /usr/bin/ruby --encoding=utf-8
```
mvidner: This means to set [`Encoding.default_external=`](https://www.rubydoc.info/stdlib/core/2.1.0/Encoding#default_external=-class_method), and it ignores any `LANG` or `LC_*` environment variables.
It's a pretty big gun. It basically shoots down https://fate.suse.com/321409 "Make YaST work in non-Unicode locales"

We could use a slightly smaller and more complicated gun:
```rb
if Encoding.default_external == Encoding::US_ASCII
  Encoding.default_external = Encoding::UTF_8
end
```

package/yast2-ruby-bindings.changes
```diff
@@ -1,3 +1,10 @@
+-------------------------------------------------------------------
+Tue Oct 23 09:52:53 UTC 2018 - jreidinger@suse.com
+
+- Fix crash caused by reading external files in ASCII encoding
```
mvidner: That's the immediate cause of the crash. But why is that?
Is glibc-locale.rpm missing? (no)
What is the ENV locale setting?
Why did that happen when the user installed in German?
Anyway, we should mention the more important thing:
> Fix encoding-related problems by assuming that file contents is always UTF-8.

src/y2start/y2start
jreidinger: but it is discouraged in Encoding module itself. And for me now, UTF-8 is safer option, as even when you have non-UTF-8 locales, you can have UTF-8 file encoded file there e.g. downloaded from network.

jreidinger: and BTW it does not help in situation like now when locale is different then encoding of file. As now you have C locale and UTF-8 file, so I worry this assumption is too narrow.

mvidner: > but it is discouraged in Encoding module itself. 

I know. The docs says
> You should not set ::default_external in ruby code as strings created before changing the value may have a different encoding from strings created after the value was changed., instead you should use ruby -E to invoke ruby with the correct default_external.

So we would avoid that problem by assigning as the very first thing in y2start.

> And for me now, UTF-8 is safer option, as even when you have non-UTF-8 locales, you can have UTF-8 file encoded file there e.g. downloaded from network.

Well, data from the network is a special case that should get an explicit encoding when it's received.

I think we are ignoring the cause of the bug report and treating the symptom with a too big hammer.

jreidinger: well, it is not too big hammer, as ruby bindings already force UTF-8 at many places. See e.g. https://github.com/yast/yast-ruby-bindings/blob/master/src/binary/Y2RubyUtils.cc#L67
so any string passed between components is UTF-8.
So this just ensure everything is UTF-8 even when reading in ruby some value from file.

mvidner: OK

PR #222, opened 2018-10-26
Fix y2start to return the client return value

mvidner:
so that the EXIT STATUS section of `man yast` is again correct

found in https://github.com/yast/yast-network/pull/686


mvidner:
oops this fix is still wrong
Anyway, to emulate the former behavior, see the old conversion code:
https://github.com/yast/yast-core/blob/a0fefb49b7497a21624fd6f65a35de56b2dbe74a/liby2/src/genericfrontend.cc#L782-L800

jreidinger:
yes, old exit code behavior is much more complex and if you want to implement it, then please in https://github.com/yast/yast-ruby-bindings/blob/master/src/ruby/yast/y2start_helpers.rb so it can be properly tested.

jreidinger:
replaced by #236

PR #223, opened 2018-12-18
[RFC] Configure global PATH

joseivanlopez:
PBI: https://trello.com/c/GMLR8Jjt/527-define-a-general-secure-path-for-all-yast-code


src/ruby/yast.rb
```diff
@@ -45,6 +45,8 @@
 require "yast/ui_shortcuts"
 require "yast/wfm"
 
+Yast.config_env
```
jreidinger: please document why we are doing it.

src/ruby/yast/yast.rb
```diff
@@ -215,4 +215,21 @@ def self.import(mname)
...
+  # Configure global environment for YaST
+  #
+  # Currently it only sets values for $PATH.
+  #
+  # By configuring $PATH, it ensures that correct external programs are executed when
+  # relative paths are given, so possible CVEs are avoided when running YaST. For example:
+  #
+  #   $ export PATH=./:$PATH
+  #   $ ruby -e "system('my_program')"                  # this could execute ./my_program
+  #   $ ruby -e "require 'yast'; system('my_program')"  # this only executes 'my_program' from safe paths
```
jreidinger: to be honest I do not much like this magic that simple require of yast will modify PATH which can be intentionally different. I think it is not job of library and instead really job of initial master process. So I would expect it at https://github.com/yast/yast-ruby-bindings/blob/master/src/y2start/y2start#L17 and this method at https://github.com/yast/yast-ruby-bindings/blob/master/src/ruby/yast/y2start_helpers.rb

This way when yast is started it sets proper ENV, but not if someone outside just require "yast" ruby bindings.

PR #224, opened 2019-01-04
[WIP] Don't crash on NoMethodError when Perl is involved

mvidner:
- https://trello.com/c/OlV7gSMM
- https://bugzilla.suse.com/show_bug.cgi?id=1119690#c9

When using the YCP component system, an exception can make YaST segfault.

This removes the segfault and replaces it with a `nil` in `y2log` :-/
Further fixes, to propagate exceptions across Perl-Ruby border crossings, are desirable.

TODO: retarget this to SLE15-GA?


src/ruby/yast/y2_exception.rb
```diff
@@ -0,0 +1,15 @@
...
+
+module Yast
+  # For passing exceptions through liby2
+  class Y2ExceptionClass < Module
+    def main
+      @exception = {}
+    end
+
+    publish variable: :exception, type: "map"
+  end
```
jreidinger, line 11: In general change looks OK for me, I just miss more documentation in this module. Especially use cases ( exceptions when calling ruby code from non-ruby parts (includes e.g. also pkg bindings) ), how that variable exception looks like and who can set it. And maybe a paragraph in readme would make sense to point to this.

jreidinger:
lets convert it to draft as it is still WIP

PR #225, opened 2019-01-22
Support for FastGettext 2.0 (bsc#1121865)

lslezak:
## Summary

- Support for FastGettext 2.0 ([bsc#1121865](https://bugzilla.suse.com/show_bug.cgi?id=1121865))
- Backward compatible with FastGettext 1.6 (present in SLE15/Leap 15.0)
- Tested manually, added an integration test (see below)
- No version upgrade, this will be released with the next maintenance update
- Added `rubygem-fast_gettext < 3.0` dependencies to avoid similar issue in the future

### Manual Tests

<details>
 <summary>Simple testing script which prints the FastGettext version and prints a translated string from YaST.</summary>

```ruby
require "yast"
include Yast::I18n
puts "FastGettext version: #{FastGettext::VERSION}"
textdomain "base"
msg = "Updating system configuration..."
puts "Message:     #{msg}"
puts "Translation: #{_(msg)}"
```

</details>

<br>

When running the script with the original Leap 15.0 package it prints this (in the Czech locale):

```
FastGettext version: 1.6.0
Message:     Updating system configuration...
Translation: Probíhá aktualizace konfigurace systému...
```

After updating the package to version 2.0 from the [devel:languages:ruby:extensions](https://build.opensuse.org/project/show/devel:languages:ruby:extensions) it prints this:

```
FastGettext version: 2.0.0
Message:     Updating system configuration...
Translation: Probíhá aktualizace konfigurace systému...
```

So it works correctly with both FastGettext versions.

### Integration Tests

We already have some unit tests for translation in [tests/i18n_spec.rb](https://github.com/yast/yast-ruby-bindings/blob/master/tests/i18n_spec.rb#L39-L44), but these tests basically mock everything so they do not test that the translation process really work.

Therefore I have added a an integration test with some real translations (provided in the attached po/mo files).


lslezak:
This is for SLE15-GA, we agreed to release it with the next real fix as we do not expect that SLE15/Leap15.0 used would use some external newer gems.

For TW (master) I'll release a new version.

archon810:
Sorry, how can I get this fix on OpenSUSE 15.0? zypper update shows no updates currently, and yast is still failing like here https://bugzilla.opensuse.org/show_bug.cgi?id=1121865.

Thanks.

PR #227, opened 2019-03-01
Fit long internal error to the screen

mchf:
Depends on https://github.com/yast/yast-yast2/pull/899

When an internal error with long backtrace is raised, it gets easily out of popup:
long_internal_error_before
With support for long text it is slightly better:
long_internal_error_after

### DIFF
--- a/src/ruby/yast/wfm.rb
+++ b/src/ruby/yast/wfm.rb
@@ -216,7 +216,7 @@ module Yast
     # @return [String] human readable exception description
     private_class_method def self.internal_error_msg(e)
       msg = "Internal error. Please report a bug report with logs.\n" \
-        "Run save_y2logs to get complete logs.\n"
+        "Run save_y2logs to get complete logs.\n\n"
 
       if e.is_a?(ArgumentError) && e.message =~ /invalid byte sequence in UTF-8/
         msg += "A string was encountered that is not valid in UTF-8.\n" \
@@ -224,7 +224,7 @@ module Yast
                "Refer to https://www.suse.com/support/kb/doc?id=7018056.\n\n"
       end
 
-      msg + "Details: #{e.message}\n" \
+      msg + "Details: #{e.message}\n\n" \
             "Caller:  #{e.backtrace.first}"
     end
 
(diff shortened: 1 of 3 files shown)


kobliha:
Please put "Caller: ..." before details and make the window bigger.

Use-case: openQA, and other screenshots: we often don't have the logs, but we get the message from the reporter. Caller is important, often even more than details, but it's not visible without scrolling here.

jreidinger:
@kobliha few issues with caller, it is much longer and it also just point to place and not to details in exception.

src/ruby/yast/wfm.rb
```diff
@@ -273,7 +273,7 @@ def self.call_builtin_wrapper(*args)
         end
       else
         Yast.import "Report"
-        Report.Error(msg)
+        Report.LongError(msg.gsub(/\n/, '<br />')
```
jreidinger: Well, when I see it, why not use new Yast2::Popup? It has stuff like details, rich text support and more.

jreidinger:
ah, we use just first line of caller. In such case it should be OK, but in general message is quite short, just in this case it is recursive failure which I intentionally change this way as previously it report only that import failed, but not root of failure.

src/ruby/yast/wfm.rb
mchf: this is in the end turned into ```Popup.LongError``` but Report adds stuff logging and some other stuff

jreidinger: and whats more also special autoyast configuration. So lets keep it as it is.

mchf:
After some tuning:
internal_error_after_2

src/ruby/yast/wfm.rb
```diff
@@ -273,7 +273,12 @@ def self.call_builtin_wrapper(*args)
         end
       else
         Yast.import "Report"
-        Report.Error(msg)
+        # Pure approximation here
+        # 50 is for usable text area width, +6 is for additional lines like
+        # button line, Error caption and so. Whole dialog is at most 20 lines
+        # high to fit into screen
+        height = [msg.size / 50 + 6, 20].min
```
jreidinger: Hmm, how this will work with 80x24 terminal? Can you try it? Maybe you can query Display size also for reasonable maximum. But also please keep code KISS as it is last stand of defense when it goes bad.

lslezak: JFYI: `String#size` is not the same as width on the screen! That does not work esp in CJK languages: `ruby -e 'puts "日".size'` => 1, but actually the character has with 2 columns (in text mode it is wide as 2 characters).

But to make it simple let's keep it, the only disadvantage is that the user will have to scroll a bit more.

PR #228, opened 2019-03-01
_() should always return a frozen string

lslezak:
*This is a proposal for the fix discussed on yast-devel@ ML, lets continue here.*

## The Problem

The `_()` translation function returns either a frozen String (from FastGettext when the translation is available) or a normal (not frozen) String when the translation is not found.

This makes the code fragile and difficult to debug as `_("foo") << _("bar")` might crash or might not depending whether the translation is available.

The problem is that when running the unit tests in Travis, Jenkins or in OBS no translation is installed and the tests run in the English locale. However, on user systems or during installation the translations are usually available. That means it's hard to test it in the same environment, the unit tests might not find this problem in the code.

Moreover the translations for new texts might be available later, so it could work fine but crash later after updating the translation package without changing the YaST module at all. Or it might crash after switching the language. That's a debugging nightmare...

### Translation Available

```console
> rpm -q yast2-trans-cs
yast2-trans-cs-84.87.20180514.157a0650d-lp150.1.1.noarch
> LC_ALL=cs_CZ.UTF-8 ruby -ryast -e 'include Yast::I18n; textdomain "base"; puts _("&Yes"); puts _("&Yes").frozen?'
&Ano
true
```

The result is a *frozen* translated string.

### Translation Not Available

```console
> rpm -q yast2-trans-es
package yast2-trans-es is not installed
> LC_ALL=es_ES.UTF-8 ruby -ryast -e 'include Yast::I18n; textdomain "base"; puts _("&Yes"); puts _("&Yes").frozen?'
&Yes
false
```

In this case you'll get the original untranslated and *not frozen* string.

## Solution

    found ? Translation._(str) : str.freeze

The fast_gettext gem returns nil when the translation is not available. In that case we return the original text but frozen to behave the same way.

## Disadvantage

The disadvantage is that freezing the String will have an unexpected side effect:

```
foo = N_("foo: ")
puts _(foo)
foo << "bar" # this will crash after freezing foo
```

But as this is just a theoretical and an obscure example which can be fixed by `puts _(foo.dup)` workaround I'm fine with the proposed change.

## Testing

I have tested the proposed patch in a SLE15-SP1 installation and everything was fine.

### DIFF
--- a/src/ruby/yast/i18n.rb
+++ b/src/ruby/yast/i18n.rb
@@ -43,12 +43,19 @@ module Yast
     end
 
     # translates given string
+    # @param str [String] the string to translate
+    # @return [String] the translated string, if the translation is not found then
+    #   the original text is returned. **The returned String is frozen!**
+    # @note **⚠ The translated string is frozen and cannot be modified. To provide
+    #   consistent results the original (not translated) string is also frozen.
+    #   This means this function modifies the passed argument! If you do not want this
+    #   behavior then pass a duplicate, e.g. `_(text.dup)`. ⚠**
     def _(str)
       # no textdomain configured yet
       if !@my_textdomain
         Yast.y2warning("No textdomain configured, cannot translate #{str.inspect}")
         Yast.y2warning("Called from: #{::Kernel.caller(1).first}")
-        return str
+        return str.freeze
       end
 
       found = true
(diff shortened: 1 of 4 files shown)


lslezak:
Any question? Comments?

src/ruby/yast/i18n.rb
```diff
@@ -60,7 +60,7 @@ def _(str)
           key_exist?(str)
         end
       end
-      found ? Translation._(str) : str
+      found ? Translation._(str) : str.freeze
```
jreidinger: for consistency I would like to see freeze also on line 51

jreidinger: and probably also do it for other methods like `n_`

In general I do not mind dup or without dup. Just please document it.

PR #230, opened 2019-05-16
A missing `textdomain` declaration is optionally an exception

mvidner:
In https://bugzilla.suse.com/show_bug.cgi?id=1130822 we have untranslated text in a dialog, even though the translators did work that particular string. The reason is a missing `textdomain` declaration in a class.

This kind of error is a programmer error which should be caught very early during the development or testing, so users should not see them. 
BUT we may have allowed this kind of bugs to creep in over time so it would be a nuisance to start failing in users faces now.
-> Make it opt-in for testing.

The exception is enabled by export Y2STRICTTEXTDOMAIN=1

no-textdomain

### DIFF
--- a/src/ruby/yast/i18n.rb
+++ b/src/ruby/yast/i18n.rb
@@ -53,8 +53,10 @@ module Yast
     def _(str)
       # no textdomain configured yet
       if !@my_textdomain
-        Yast.y2warning("No textdomain configured, cannot translate #{str.inspect}")
-        Yast.y2warning("Called from: #{::Kernel.caller(1).first}")
+        msg = "No textdomain configured in #{self.class}, " \
+              "cannot translate #{str.inspect}"
+        raise msg if ENV["Y2STRICTTEXTDOMAIN"]
+        Yast.y2warning(1, "%1", msg) # skip 1 frame
         return str.freeze
       end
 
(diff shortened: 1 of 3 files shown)


src/ruby/yast/i18n.rb
```diff
@@ -53,9 +53,14 @@ def textdomain(domain) # usually without brackets like textdomain "example"
     def _(str)
       # no textdomain configured yet
       if !@my_textdomain
-        Yast.y2warning("No textdomain configured, cannot translate #{str.inspect}")
-        Yast.y2warning("Called from: #{::Kernel.caller(1).first}")
-        return str.freeze
+        begin
+          raise "No textdomain configured in #{self.class}, " \
```
lslezak: Um, I do not like this begin - raise - rescue - raise structure at all. :worried: 

To avoid repeating the message I'd prefer something like this:

```ruby
msg = "No textdomain configured in #{self.class}, " \
"cannot translate #{str.inspect}"

raise msg if ENV["Y2STRICTTEXTDOMAIN"]

Yast.y2warning(1, "%1", msg)
return str.freeze
```

mvidner: You are right, this is better, thanks!

mvidner:
@rwx788 Rodion, I would like to enable this in openQA tests across all of YaST (post-SP1) by setting the environment variable `Y2STRICTTEXTDOMAIN=1`. Does it make sense? It it easy to do?

mvidner:
Enabling this in RSpec tests will also help, let's find a good place to put it in

rwx788:
> @rwx788 Rodion, I would like to enable this in openQA tests across all of YaST (post-SP1) by setting the environment variable `Y2STRICTTEXTDOMAIN=1`. Does it make sense? It it easy to do?

That totally makes sense. We already parse logs for such errors, and we have one scenario specifically dedicated to this case, to click more menus, etc. I'll create backlog item for it. Thanks!

mvidner:
@rwx788 just a reminder that this would catch some untranslated texts, but I don't know where to enable it, and what is a good time in the release cycle for it

rwx788:
> @rwx788 just a reminder that this would catch some untranslated texts, but I don't know where to enable it, and what is a good time in the release cycle for it

Hi @mvidner! We have already enabled `Y2STRICTTEXTDOMAIN` in couple of scenarios. Do you think it makes sense to enable in all the tests by default?

PR #231, opened 2019-05-20
Avoid breaking the build of yast2.rpm (bsc#1130822)

mvidner:
A fixup for #230 - https://bugzilla.suse.com/show_bug.cgi?id=1130822#c13

If /usr/share/YaST2/locale does not exist (eg. during the building of yast2.rpm) FastGettext will complain. So in that case bail out of `textdomain` and `_`+friends.
    
Here's how it broke without this fix:
    
+Log	No textdomain configured in Yast::CommandLineClass, cannot translate "Print the help for this module"

The log line *appears* because it is now reported at CommandLine.rb
(previously i18n.rb) which matches the "testedfiles" filter of the
testsuite.

The error *happens* the "textdomain" method returns early before
remembering its argument.

### DIFF
--- a/src/ruby/yast/i18n.rb
+++ b/src/ruby/yast/i18n.rb
@@ -53,10 +53,12 @@ module Yast
     def _(str)
       # no textdomain configured yet
       if !@my_textdomain
-        msg = "No textdomain configured in #{self.class}, " \
-              "cannot translate #{str.inspect}"
-        raise msg if ENV["Y2STRICTTEXTDOMAIN"]
-        Yast.y2warning(1, "%1", msg) # skip 1 frame
+        if File.exist?(LOCALE_DIR)
+          msg = "No textdomain configured in #{self.class}, " \
+                "cannot translate #{str.inspect}"
+          raise msg if ENV["Y2STRICTTEXTDOMAIN"]
+          Yast.y2warning(1, "%1", msg) # skip 1 frame
+        end
         return str.freeze
       end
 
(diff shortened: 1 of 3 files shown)


src/ruby/yast/i18n.rb
```diff
@@ -1,3 +1,4 @@
+# coding: utf-8
```
lslezak: NP: IIRC the UTF-8 encoding is the default in Ruby 2.0+, this line is not need. Or did I overlok something?

jreidinger: no, really no longer needed for newer ruby

mvidner: My Emacs mode let this slip through, I'll fix that

mvidner:
Still no good. With this change, yast2.rpm fails with

```
[   24s]   1) #<Yast::IPClass:0x000055d850de8180> #Valid4 must return valid IPv4 description
[   24s]      Failure/Error: expect(subject.Valid4).not_to be_empty
[   24s] 
[   24s]      FastGettext::Storage::NoTextDomainConfigured:
[   24s]        Current textdomain ("base") was not added, use FastGettext.add_text_domain !
[   24s]      # /home/abuild/rpmbuild/BUILD/yast2-4.2.4/library/types/src/modules/IP.rb:56:in `Valid4'
[   24s]      # ./ip_test.rb:12:in `block (3 levels) in <top (required)>'
```

FWIW, the original change introducing the short circuit was 9a411e3cb08f4e5222c1c77385cbc051dd0382d1

PR #233, opened 2019-09-23
Add new ItemSelector to the UI Interpreter

shundhammer:
## Trello

https://trello.com/c/7qI1AAvN/1300-5-libyui-1-of-n-selection-widget-ysingleitemselector-ymultiitemselector

## Related PR

https://github.com/libyui/libyui/pull/148

## Description

It's not enough to add a new widget to the YCPDialogParser; it also has to be registered with our Ruby bindings here.


package/yast2-ruby-bindings.spec
```diff
@@ -41,8 +41,8 @@ BuildRequires:  ruby-devel
 Requires:       yast2-core >= 3.2.2
 BuildRequires:  yast2-core-devel >= 3.2.2
 # UI.SetApplicationTitle
-Requires:       yast2-ycp-ui-bindings       >= 3.2.0
-BuildRequires:  yast2-ycp-ui-bindings-devel >= 3.2.0
+Requires:       yast2-ycp-ui-bindings       >= 4.3.0
```
jreidinger: In general it looks good, I am just a bit confused here that 4.2.X depends on 4.3.X . What is planned versioning?

shundhammer: libyui and related packages have their own versioning scheme. It has gone quite out of sync with the other YaST packages (since quite a while ago).

The last yast2-ycp-ui-bindings was 4.2.0; since this now supports a whole new widget, we decided to bump the minor version number for all UI packages,i.e. libyui, libyui-qt, libyui-ncurses, and consequently also yast-ycp-ui-bindings.

jreidinger: then we really need to document that yast2-ycp-ui-bindings does not follow yast versioning conventions, as other yast modules do follow it up. For libyui it is fine for me, but yast2 bindings it is a bit confusing.

shundhammer: I really don't care one way or the other. But the sheer number of packages that need changes is driving me over the edge here. Who came up with the idea to split that up in so many tiny packages in the first place?

shundhammer: Now bumping yast2-ycp-ui-bindings only to version 4.2.2.

PR #235, opened 2019-12-05
 S390: Evaluating an architecture specific string which will be shown…

schubi2:
… in the title bar (ncurses) or in the banner (qt)

### DIFF
--- a/src/ruby/yast/y2start_helpers.rb
+++ b/src/ruby/yast/y2start_helpers.rb
@@ -1,4 +1,3 @@
-
 module Yast
   module Y2StartHelpers
     # Configure global environment for YaST
--- a/src/ruby/yast/y2start_helpers.rb
+++ b/src/ruby/yast/y2start_helpers.rb
@@ -79,14 +79,12 @@ module Yast
       if is_s390
         # e.g. stdout "2964 = z13 IBM z13" transfered into "IBM z13"
         arch_array = read_values.split("=")
-        if arch_array.size > 1
-          architecture = arch_array[1].strip
-        else
-          architecture = arch_array[0]
-        end
+        arch_array.shift if arch_array.size > 1
+        architecture = arch_array.join(' ').strip
         arch_array = architecture.split(' ')
         arch_array.shift if arch_array.size > 1
         architecture = arch_array.join(' ')
+
         if !Yast::UI.TextMode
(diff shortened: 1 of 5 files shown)


src/y2start/y2start
```diff
@@ -48,16 +48,13 @@ NO_CLI_CLIENTS = ["installation", "view_anymsg", "firstboot", "scc"].freeze
 Yast.ui_create(args[:server_name], args[:server_options])
 # set application title bsc#1033161
 Yast.import "UI"
-# do not fail if gethostname failed
-hostname = Socket.gethostname rescue ""
-hostname = "" if hostname == "(none)"
-hostname = " @ #{hostname}" unless hostname.empty?
+Yast.import "Arch"
```
jreidinger: I am not sure if this is good idea to start more depending in starting script on YaST itself. I vote in this case for some easier and more straighforward way to detect s390.
I know here it import UI, but only to do initial initialization of title and we already have a lot of problems with it ( see below ).

src/ruby/yast/y2start_helpers.rb
```diff
@@ -69,6 +71,33 @@ def self.setup_signals
...
+          # The environment variable YAST_BANNER will be read and shown
+          # in libyui-qt.
+          ENV["YAST_BANNER"] = architecture
+          architecture = ""
+        end
+      else
+        architecture = ""
+      end
+      left_title = "YaST2 - #{client_name}#{hostname}"
+      left_title + architecture.rjust(80-left_title.size)
```
jreidinger: test case? this is critical part of YaST so I would like to see it very well test covered.

src/y2start/y2start
schubi2: What are you suggesting ?

jreidinger: something like calling `arch` and check output?

schubi2: I will take the "arch" call instead. Is it OK to use Yast::Execute.on_target at least ? ;-)

schubi2: I see it also depend on YaST. So I will change it too.

jreidinger: well, Execute is basically cheetah with some flavor and flavor is mostly logger. So it can cause issue if we want to tune a bit logger in this script. Potential issue is to use `on_target` without `!` which can open popup.

PR #236, opened 2019-12-17
Exit codes

jreidinger:
Implement proper exit codes of yast program

It is currently identical behavior as in original y2core.
References:
https://github.com/yast/yast-core/blob/master/liby2/src/genericfrontend.cc#L777
https://github.com/yast/yast-core/blob/master/liby2/src/include/y2/exitcodes.h#L25
https://github.com/yast/yast-core/blob/master/liby2/src/genericfrontend.cc#L115

replaces https://github.com/yast/yast-ruby-bindings/pull/222

### DIFF
--- a/package/yast2-ruby-bindings.changes
+++ b/package/yast2-ruby-bindings.changes
@@ -1,4 +1,12 @@
 -------------------------------------------------------------------
+Tue Dec 17 14:48:01 UTC 2019 - Josef Reidinger <jreidinger@suse.com>
+
+- Fix returning error codes from y2start start point. Helps with
+  CLI exit codes and also with failed installation (helps e.g. with
+  bsc#1144351 and yast lan CLI)
+- 4.2.6
+
+-------------------------------------------------------------------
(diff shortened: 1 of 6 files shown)


mvidner:
It looks OK but the Travis failure is messy, I can't tell whether this PR is causing it.

jreidinger:
@mvidner me neither and it is strange that only push one and not pr one failed. Restarting job to see if it is just temporary issue.

jreidinger:
@mvidner with additional debug output it now prints segfault. Sadly I cannot reproduce it locally, as `rake osc:build` skips this test ( so also BS is not affected ). Not sure what case it as I do not have core dump, but this change does not look like reason for segfault ( which is probably race condition ).

mvidner:
> Sadly I cannot reproduce it locally, as `rake osc:build` skips this test

It skips because tmux is not installed. That can be solved with `rake osc:build["-x tmux"]`... and then it passes for me. :woman_shrugging:

mvidner:
I tried various ways to reproduce the test failure locally but it always succeeded. So I restarted the Travis job and now it is green.

PR #237, opened 2020-01-03
Another exit codes

jreidinger:
This is a fix-up of #236 and Josef should really learn to write at least minimal PR descriptions! Hint hint!

### DIFF
--- a/src/ruby/yast/wfm.rb
+++ b/src/ruby/yast/wfm.rb
@@ -331,6 +331,9 @@ module Yast
         handle_abort_exception(e, client)
         exit
       rescue Exception => e
+        # Don't interfere with RSpec, such as RSpec::Mocks::MockExpectationError
+        raise e if e.class.to_s.start_with?("RSpec::")
+
         handle_exception(e, client)
         false
       end
--- a/package/yast2-ruby-bindings.changes
+++ b/package/yast2-ruby-bindings.changes
@@ -1,4 +1,11 @@
 -------------------------------------------------------------------
+Fri Jan  3 15:02:24 UTC 2020 - Josef Reidinger <jreidinger@suse.com>
+
+- Fix previous fix to really set exit code in ruby wrapper
+  (bsc#1144351)
+- 4.2.7
+
+-------------------------------------------------------------------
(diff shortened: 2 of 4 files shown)


mvidner:
I think we can add an integration test for this, adding an `ADD_TEST` to `tests/CMakeLists.txt`. A prototype:

```console
$ cat /tmp/a.rb 
42
$ RUBYLIB=src/ruby src/y2start/y2start /tmp/a.rb testsuite; echo $?
58
```

jreidinger:
@mvidner OK, I try to do it and I have test that works locally. But it does not work automatic. Reason? It needs installed yast2-ruby-bindings to run properly. Because y2start basically runs component broker for running client and component system does not know about ruby clients without ruby bindings installed.
So I can add test, but it cannot be run automatically.

Do you think it is worth having it? or should I merge it without that test?

mvidner:
Oh, it's a bootstrap problem :-(

I'd be OK with deferring tests to openQA, but when I look at the bug you mention, https://bugzilla.suse.com/show_bug.cgi?id=1144351 , I don't see an exit code test. Do we have one elsewhere?

jreidinger:
@mvidner I do not see explicit one, but I found it during testing some bugs for CLI where bug reporter for older distribution report that it returns 16 and for me it return 0 and I am trying to find why.

src/ruby/yast/wfm.rb
```diff
@@ -331,6 +331,9 @@ def self.run_client(client)
         handle_abort_exception(e, client)
         exit
       rescue Exception => e
+        # Don't interfere with RSpec, such as RSpec::Mocks::MockExpectationError
+        raise e if e.class.to_s.start_with?("RSpec::")
```
mvidner: @jreidinger this is the fix mentioned in https://github.com/yast/yast-yast2/pull/1004

mvidner:
OK, so now we should be able to add a simple openQA test where `yast2 $module nosuchcommand` really produces a non-zero exit code.

src/ruby/yast/wfm.rb
jreidinger: can be written as `raise if e.class.to_s.start_with?("RSpec::")`

PR #239, opened 2020-02-26
Fixes for gensalt handling with libxcrypt.

besser82:
The `libxcrypt` library in version >= 4 offers support for safely generating entropy itself.  Besides that, the `<crypt.h>` header of the library already defines `CRYPT_GENSALT_OUTPUT_SIZE`.


jreidinger:
follow up of this PR https://github.com/yast/yast-ruby-bindings/pull/250

PR #241, opened 2020-04-16
Ruby27

jreidinger:
waiting for branching of SP2

bugzilla: https://bugzilla.suse.com/show_bug.cgi?id=1169442#c2


jreidinger:
@mvidner yes, I use it a lot for testing. Thanks

PR #243, opened 2020-06-04
set also script name

jreidinger:
see https://bugzilla.suse.com/show_bug.cgi?id=1172275#c18

intentionally no changes as it goes as part of previous fix. Catched by Maintenance QA


dgdavid:
BTW, is it related to https://github.com/yast/yast-ruby-bindings/pull/242?

PR #244, opened 2020-06-25
Backport to SLE-12-SP4: Fix Gem Loading

shundhammer:
## Bugzilla

https://bugzilla.suse.com/show_bug.cgi?id=1172848

## Trello

https://trello.com/c/tZ9MKejZ/

## Problem

An appliance made with SUSE Studio (i.e. using KIWI) based on SLE-12-SP4 fails since the last Ruby security update in their yast-firstboot step where they do various configurations after extracting their image.

It turned out that they are invoking other YaST modules (e.g. network) as external processes by invoking `y2base` directly.

## Underlying Problems

Ruby does not really have an API for embedding; to integrate it into YaST, we really hammered it into shape to make it work. That involved loading some Ruby gems early upon YaST startup: E.g. the `fast_gettext` gem which we use for translations. How this is done tends to differ slightly between individual Ruby versions, so it has to be adapted when that happens.

## Solution

- Backported some changes that @jreidinger had already made for SLE-12-SP2: https://github.com/yast/yast-ruby-bindings/pull/242, https://github.com/yast/yast-ruby-bindings/pull/243. This is what this PR is all about.

- Recommended to the team who makes that appliance to use the `/sbin/yast` script because this is less fragile; some initializations are done there. Calling `y2base` is subtly different; or call the Ruby code directly from Ruby and not as an external process.

### Without Patches: Fail

y2base-vs-y2start

Invoking `y2base` makes the called YaST module fail with exit code 16.
Invoking the same with `yast2` works fine.

### With Patches: Success

with-backport

With the version from this PR, it works even with `y2base`.

## Version number

The source code for those ruby-bindings has not changed at all between SLE-12-SP4, so both will get version 3.2.17.

We discussed this in IRC, and both @lslezak and @jreidinger agree.

## Related PR

This will also go to SLE-12-SP5.

### DIFF
--- a/package/yast2-ruby-bindings.changes
+++ b/package/yast2-ruby-bindings.changes
@@ -1,4 +1,13 @@
 -------------------------------------------------------------------
+Thu Jun 25 12:40:48 UTC 2020 - Stefan Hundhammer <shundhammer@suse.com>
+
+- Backported jreidinger's patches to SLE-12-SP4 (bsc#1172848):
+  - PR#242: Fix enc loading
+    (fix gems loading with updated ruby (bsc#1172275))
+  - PR#243: set also script name
+- 3.2.17
+
+-------------------------------------------------------------------
 Tue Oct 23 09:52:53 UTC 2018 - jreidinger@suse.com
 
 - Fix encoding-related problems by assuming that file contents is
--- a/package/yast2-ruby-bindings.spec
+++ b/package/yast2-ruby-bindings.spec
@@ -17,7 +17,7 @@
 
 
 Name:           yast2-ruby-bindings
-Version:        3.2.16
+Version:        3.2.17
 Url:            https://github.com/yast/yast-ruby-bindings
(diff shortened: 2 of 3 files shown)


shundhammer:
The original PRs went to SLE-12-SP2 that obviously had older Ruby versions, so I don't see a problem there.

jreidinger:
yes, this change works also with older ruby.

package/yast2-ruby-bindings.changes
```diff
@@ -1,3 +1,12 @@
+-------------------------------------------------------------------
+Thu Jun 25 12:40:48 UTC 2020 - Stefan Hundhammer <shundhammer@suse.com>
+
+- Backported jreidinger's patches to SLE-12-SP4 (bsc#1172848):
```
jreidinger, line 4: just nitpick, it is not backport, but forward port, as original patches was for SLE12 SP2

PR #249, opened 2020-09-17

mvidner:
- https://trello.com/c/Vg5AS4wI
- https://bugzilla.suse.com/show_bug.cgi?id=1174198

Snímek obrazovky_2020-09-17_cfa-error

### DIFF
--- a/package/yast2-ruby-bindings.changes
+++ b/package/yast2-ruby-bindings.changes
@@ -1,4 +1,11 @@
 -------------------------------------------------------------------
+Thu Oct 29 11:30:49 UTC 2020 - schubi <schubi@localhost>
+
+- Reporting augeas parsing errors and displaying them in rich-text
+  format (bsc#1174198).
+- 4.3.6
+
+-------------------------------------------------------------------
 Tue Oct 13 14:42:52 UTC 2020 - Martin Vidner <mvidner@suse.com>
 
 - Add automatic TUI (ncurses) tests using tmux (bsc#1165388).
--- a/package/yast2-ruby-bindings.changes
+++ b/package/yast2-ruby-bindings.changes
@@ -1,5 +1,5 @@
 -------------------------------------------------------------------
-Thu Oct 29 11:30:49 UTC 2020 - schubi <schubi@localhost>
+Thu Oct 29 11:30:49 UTC 2020 - Stefan Schubert <schubi@suse.com>
 
 - Reporting augeas parsing errors and displaying them in rich-text
   format (bsc#1174198).
--- a/src/ruby/yast/wfm.rb
+++ b/src/ruby/yast/wfm.rb
@@ -211,13 +211,14 @@ module Yast
       !Mode.auto && !Debugger.unwanted? && Debugger.installed?
     end
 
+    private_class_method def self.escape_angle_brackets(str)
+      ret.gsub!(/</, "&lt;")
+      ret.gsub(/>/, "&gt;")
+    end
+
     # @param [CFA::AugeasParsingError] e the caught exception
     # @return [String] human readable exception description
     private_class_method def self.parsing_error_msg(e)
-      augeas_message = e.message.dup
-      augeas_message.gsub!(/</, "&lt;")
-      augeas_message.gsub!(/>/, "&gt;")
-
       msg = "Parse error while reading file #{e.file}<br>" \
             "YaST cannot continue and will quit.<br>" \
             "<br>" \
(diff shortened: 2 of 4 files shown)


src/ruby/yast/wfm.rb
```diff
@@ -212,6 +212,24 @@ def self.call_builtin_wrapper(*args)
...
 
+    # @param [CFA::AugeasParsingError] e the caught exception
+    # @return [String] human readable exception description
+    private_class_method def self.parsing_error_msg(e)
+      msg = "Parse error while reading file #{e.file}\n" \
+            "YaST cannot continue and will quit.\n" \
+            "\n" \
+            "Possible causes and remedies:\n" \
+            "1. You made a mistake when changing the file by hand,\n" \
+            "   the syntax is invalid. Try reverting the changes.\n" \
```
jreidinger, line 223: as generic top level catch it is OK, bu still I do not like possible solution and outcome for users. I prefer way how it is done in bootloader, when this error happen it allows to repropose from scratch to get again working configuration - https://github.com/yast/yast-bootloader/blob/master/src/lib/bootloader/config_dialog.rb#L38

mvidner: OK. So you're saying that in addition to this, each user-facing module should have its own General Rescue, with proposing a working configuration from scratch as a possible option.

For the bootloader I see it as especially important since, well, can't boot, can't fix anything.
@mchf @teclator is there a similar option for the network module already?

jreidinger: @mvidner well, it is not just about not booting cannot fix, but in general provide user option how to handle in user friendly way when configuration is broken for whatever reason. Or maybe we can suggest user to remove configuration file and force reinstall of package to get back default configuration and rerun yast module? not sure what is the best option from user POV

mvidner: Ah, network has already been improved in [yast/yast-network@0c717ce](https://github.com/yast/yast-network/commit/0c717ceb18e8f64d26b09f3e2136593700d8dc08#diff-a44342e8d97412ca7486bd2d61e8e429L269-R273), resulting in this:

network_error_dialog

mvidner:
When byebug is not installed (or the newly recognized Y2DEBUGGER=0 is in effect) a richtext window is used instead, showing a wrapped text:

distinguish-rich

mvidner:
The rich text will nicely wrap lines, but it will not escape angle brackets:

```console
$ cat raise_angle_brackets.rb
Class.new.no_such_method
$ /usr/sbin/yast2 ./raise_angle_brackets.rb  # question, plain text
$ Y2DEBUGGER=0 /usr/sbin/yast2 ./raise_angle_brackets.rb  # message, rich text
```

raise-question 
raise-rich

package/yast2-ruby-bindings.changes
```diff
@@ -1,3 +1,10 @@
+-------------------------------------------------------------------
+Thu Oct 29 11:30:49 UTC 2020 - schubi <schubi@localhost>
```
mvidner, line 2: real name and email please

src/ruby/yast/wfm.rb
```diff
@@ -208,35 +208,57 @@ def self.call_builtin_wrapper(*args)
 
     private_class_method def self.ask_to_run_debugger?
       Yast.import "Mode"
+      !Mode.auto && !Debugger.unwanted? && Debugger.installed?
+    end
+
+    private_class_method def self.escape_angle_brackets(str)
```
mvidner, line 214: IMHO it's better to use the stdlib method:

```rb
require "cgi"
include CGI::Escape
escapeHTML "&<>'\""  # => "&amp;&lt;&gt;&#39;&quot;"
```

src/ruby/yast/wfm.rb
```diff
@@ -208,35 +208,57 @@ def self.call_builtin_wrapper(*args)
...
 
-      !Mode.auto && Debugger.installed?
+    # @param [CFA::AugeasParsingError] e the caught exception
+    # @return [String] human readable exception description
+    private_class_method def self.parsing_error_msg(e)
+      msg = "Parse error while reading file #{e.file}<br>" \
+            "YaST cannot continue and will quit.<br>" \
+            "<br>" \
+            "Possible causes and remedies:<br>" \
+            "1. You made a mistake when changing the file by hand,<br>" \
```
mvidner, line 227: Since we use RichText, have you consodered `<ol>` and `<li>`?

src/ruby/yast/wfm.rb
```diff
@@ -208,35 +208,57 @@ def self.call_builtin_wrapper(*args)
...
+            "   the syntax is invalid. Try reverting the changes.<br>" \
+            "2. The syntax is in fact valid but YaST does not recognize it.<br>" \
```
mvidner, line 229: IMHO the line break inside the numbered items should be removed

schubi2:
We are beginning with SLES15-SP2:
https://github.com/yast/yast-ruby-bindings/pull/253
And merging this one in the next step for master.

PR #250, opened 2020-09-24
Crypt improvements

jreidinger:
follow up of https://github.com/yast/yast-ruby-bindings/pull/239
I clean this up to support only the latest libxcrypt as it is already in SLE15 SP3, Leap 15.3 and TW.
bsc: https://bugzilla.suse.com/show_bug.cgi?id=1176924

### DIFF
--- a/package/yast2-ruby-bindings.changes
+++ b/package/yast2-ruby-bindings.changes
@@ -1,5 +1,5 @@
 -------------------------------------------------------------------
-Wed Feb 26 19:46:00 UTC 2020 - besser82@fedoraproject.org
+Thu Sep 24 19:46:00 UTC 2020 - besser82@fedoraproject.org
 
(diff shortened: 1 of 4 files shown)


mvidner (changes requested):
Thanks!

Please rebase to fix the confusing 42.3 version in one commit message (15.3 is correct)
Please link to the bug in the PR description

mvidner:
BTW the ncurses test which is piggybacked here is failing with "pure virtual method called"
https://travis-ci.org/github/yast/yast-ruby-bindings/builds/730012983#L2542

jreidinger:
@mvidner yes, it is not related to this PR and I do not know why. I see it in all push build and in none pr kind. Not sure why.

mvidner (approved):

PR #251, opened 2020-10-06
automatic tests for TUI (ncurses)

mvidner:
- https://trello.com/c/MUi4zmBT
- see also https://github.com/yast/yast-ycp-ui-bindings/pull/56


tests/CMakeLists.txt
```diff
@@ -12,3 +17,8 @@ endforeach(test)
 
 ADD_TEST("integration"  ruby ${CMAKE_CURRENT_SOURCE_DIR}/integration/run.rb)
 ADD_TEST("translations" rspec --format doc ${CMAKE_CURRENT_SOURCE_DIR}/integration/translations_spec.rb)
+
+file(GLOB libyui_tests "libyui/*.test")
```
jreidinger, line 21: Well, tests are in tmux which I do not know much, but fine for me. I have other issues:

1. why it lives in yast2-ruby-bindings? it should be in libyui itself or libyui-bindings if you really want to use ruby and not write it as example program in C++. Do you really expect that libyui contributors search for it here and also run it on their magei/buntu/etc? Ideally it should be `make && make check` in libyui itself

2. It tests too late. What happens here? It needs libyui, its bindings, then ycp-ui-bindings, then yast2-core and then yast2-ruby-bindings and then you can see what failed. To be honest I think it will be in comparable speed to get result from openqa itself. It is too slow for unit testing and looks more like full integration testing, where openqa is a better as it test more and has more tests.

mvidner:
@jreidinger 

> 1. why it lives in yast2-ruby-bindings? it should be in libyui itself or
   libyui-bindings if you really want to use ruby and not write it as example
   program in C++. Do you really expect that libyui contributors search for it
   here and also run it on their magei/buntu/etc? Ideally it should be make &&
   make check in libyui itself 

The motivation for this is: we make some changes to libyui (add sort keys, add
menus, add nested table rows) and it breaks in unexpected places. We need
tests.

The deciding factor is: who writes the tests and in what language? My answer is: the yast people, in Ruby. 

We have already tried writing tests in C++, see
https://github.com/libyui/libyui-test , it hasn't taken off.


https://github.com/libyui/libyui is an abstract base, good only for elementary
unit tests. Sorry, cannot work for the tests that we need.

https://github.com/libyui/libyui-bindings is IMHO immature and unused. And
https://github.com/yast/yast-ycp-ui-bindings/tree/master/examples already has
the canonical collection of examples which we can directly use or adapt. 

Libyui contributors are, for now, a secondary concern. We first need to enable
testing for the core team.


> 2. It tests too late. What happens here? It needs libyui, its bindings, then
   ycp-ui-bindings, then yast2-core and then yast2-ruby-bindings and then you
   can see what failed. To be honest I think it will be in comparable speed to
   get result from openqa itself. It is too slow for unit testing and looks
   more like full integration testing, where openqa is a better as it test
   more and has more tests. 

Our team simply does not write openQA tests. IMHO it's YaST tests or no tests.

And no, this can still be run by the developer locally as they modify libyui. Installing a handful of yast packages on top of libyui is still much less of a burden than waiting for the massive openQA machinery.

jreidinger:
@mvidner still it does not much convince to me to be the tests that helps with development. Another idea, if tests are for ncurses only, then I suggest to place it in libyui-ncurses and make it mandatory ( so part of rpm spec build and also travis ). so we immediately see in pr if it breaks tests or not. And similar way we can use some qt testing framework for libyui-qt. I think it should live with code otherwise it become obsolete, even here in ruby-bindings, as it does not change often and you catch it too late.
For me reason why libyui-test failed is that it is not part of code. If tests are run as part of travis and ideally rpm build, then it will be maintained and provide early feedback if something goes south.

ancorgs:
> @mvidner still it does not much convince to me to be the tests that helps with development. Another idea, if tests are for ncurses only, then I suggest to place it in libyui-ncurses and make it mandatory ( so part of rpm spec build and also travis ). so we immediately see in pr if it breaks tests or not.

It's true that it would be good to have the tests triggered for every pull request opened against libyui-ncurses (adding the tests to the rpm spec build doesn't sound so urgent to me). If such pull request needs adaptation to the tests, having them separated can indeed be a problem.

On the other hand, previous @mvidner's arguments have convinced me about the convenience of having the tests in this repository. For me, ensuring the tests are seen, understood and maintained by the YaST Team (who are mainly Ruby developers that could easily come up with tests by basically copy/pasting real pieces of YaST) is the most relevant criteria here.

tests/libyui/tmux-uitest.sh
```diff
@@ -0,0 +1,89 @@
+SESSION=uitest
```
lslezak, line 1: Again, some header describing the file would be nice here...

lslezak (commented):
I'm getting this test errors when running `rake osc:build` locally:

```
[   17s] The following tests FAILED:
[   17s] 	 25 - integration (Failed)
[   17s] 	 27 - /home/abuild/rpmbuild/BUILD/yast2-ruby-bindings-4.3.5/tests/libyui/table_sort_wrong_cell_1165388.test (Failed)
[   17s] Errors while running CTest
[   17s] make: *** [Makefile:137: test] Error 8
```

How can I get more details about the failures?

mvidner:
Just when I thought I can finally finish this it turns out that latest libyui broke the single test that I have here :frowning_face:

tests/libyui/tmux-uitest.sh
```diff
@@ -0,0 +1,106 @@
...
+# $1 shell command for sh -c
+tmux_new_session() {
+    if $VERBOSE; then
+        echo Starting session
+    fi
+    # -s session name
+    # -x width -y height,
+    # -d detached
+    # FIXME: sleep to be able to see errors when running $1
+    tmux new-session -s "$SESSION" -x 80 -y 24 -d sh -c "$1; sleep 9999"
```
mvidner, line 16: @lslezak you asked "How can I get more details about the failures?"

This sleep lets us capture the screen after a command fails and exits.

Tmux has a `remain-on-exit` option but I haven't figured out how to enable it yet :-(

tests/libyui/menu_hotkeys_1177760_spec.rb
```diff
@@ -0,0 +1,36 @@
...
+    @base = "menu_hotkeys_#{bug}"
+
+    yast_ncurses = "#{__dir__}/yast_ncurses"
+    example_dir = "/usr/share/doc/packages/yast2-ycp-ui-bindings/examples"
+    @tui = TmuxTui.new_session "#{yast_ncurses} #{example_dir}/MenuBar1.rb"
+
+    ex.run
+
+    if @tui.has_session?
+      @tui.kill_session
```
jreidinger, line 15: why two methods? I would have just one like `kill_session` that internally check if session exists.

tests/libyui/menu_hotkeys_1177760_spec.rb
```diff
@@ -0,0 +1,36 @@
+require_relative "rspec_tmux_tui"
+
+describe "Menu Item" do
+  bug = "1177760"
+  around(:each) do |ex|
+    @base = "menu_hotkeys_#{bug}"
```
jreidinger, line 6: where is it used?

tests/libyui/menu_hotkeys_1177760_spec.rb
```diff
@@ -0,0 +1,36 @@
...
+
+    yast_ncurses = "#{__dir__}/yast_ncurses"
+    example_dir = "/usr/share/doc/packages/yast2-ycp-ui-bindings/examples"
+    @tui = TmuxTui.new_session "#{yast_ncurses} #{example_dir}/MenuBar1.rb"
```
jreidinger, line 10: I expect this whole creation need to be redone in each test. So why not simply have helper in TmuxTui and have something like `@tui = TmuxTui.testing_client("MenuBar1.rb")` instead of those three lines?

tests/libyui/rspec_tmux_tui.rb
```diff
@@ -0,0 +1,76 @@
...
+    @y = y
+    @detach = detach
+    @session_name = session_name || new_session_name
+
+    system "tmux", "new-session",
+           "-s", @session_name,
+           "-x", @x.to_s,
+           "-y", @y.to_s,
+           *(@detach ? ["-d"] : [] ),
+           "sh", "-c", "#{@shell_command}; sleep 9999"
```
jreidinger, line 25: is that long sleep still needed? I think you can do in rspec specific expect name like `expect(@tui).to include_in_pane("Last Event")` and have own specific error handler there if you need to somehow capture it.

tests/libyui/menu_hotkeys_1177760_spec.rb
```diff
@@ -0,0 +1,36 @@
+require_relative "rspec_tmux_tui"
+
```
lslezak, line 2: NP: It would be nice to have a full bug URL here, just having a bug number might not be enough for people not familiar with libyui or SUSE.

tests/libyui/menu_hotkeys_1177760_spec.rb
lslezak: NP: Maybe we should print a warning on STDERR in this case, each test tries to finish cleanly (by pressing "Quit" or "Close" button), if the test is still running then something went wrong.

tests/libyui/rspec_tmux_tui.rb
```diff
@@ -0,0 +1,76 @@
...
+    esc = color ? "-e" : ""
+    # FIXME: failure of the command?
+    `tmux capture-pane -t #{session_name.shellescape} -p #{esc}`
+  end
+
+  def capture_pane_to(filename)
+    txt = capture_pane(color: false)
+    esc = capture_pane(color: true)
+    File.write("#{filename}.out.txt", txt)
+    File.write("#{filename}.out.esc", esc)
```
lslezak, line 43: I'd prefer `.raw` suffix so it's clear that it's the original unprocessed screen dump.

tests/libyui/rspec_tmux_tui.rb
```diff
@@ -0,0 +1,76 @@
...
+
+  # @param keys [String] "C-X" for Ctrl-X, "M-X" for Alt-X, think "Meta";
+  #   for details see:
+  #     man tmux | less +/"^KEY BINDINGS"
+  def send_keys(keys)
+    system "tmux", "send-keys", "-t", session_name, keys
+  end
+    
+  def has_session?
+    system "tmux", "has-session", "-t", session_name
```
lslezak, line 70: Um, because of that `sleep 9999` the session will be always present, or am I wrong :question:

tests/libyui/table_sort_spec.rb
```diff
@@ -0,0 +1,47 @@
...
+      @tui.capture_pane_to("#{base}-3-first-row-activated")
+
+      expect(@tui.capture_pane).to match(/name-aaa.VALUE-AAA/)
+
+      @tui.send_keys "M-C"      # &Close
+    end
+
+    bug = "1177145"
+    it "ChangeWidget(_, :CurrentItem) activates the correct line, boo##{bug}" do
+      skip "not fixed yet"
```
lslezak, line 32: Maybe marking as `pending` would be better than skipping completely?

https://relishapp.com/rspec/rspec-core/v/3-8/docs/pending-and-skipped-examples

tests/libyui/menu_hotkeys_1177760_spec.rb
mvidner: in the screenshot filenames: `@tui.capture_pane_to("#{@base}-1-initial")`

tests/libyui/tmux-uitest.sh
mvidner: Ha! Totally intuitive for a beginner... NOT

`tmux set-hook -g session-created 'set remain-on-exit on' \; new-session sh -c ...`

Found in https://github.com/tmux/tmux/issues/787

tests/libyui/rspec_tmux_tui.rb
mvidner: Unfortunately, as I understand it, there is no such thing as a raw terminal dump, unless we're talking about the linux console devices (`/dev/vcsa7` having colors for `/dev/vcs7`). I feel `.esc` is more specific.

jreidinger (approved):

PR #252, opened 2020-10-14
Fixed if statement in yast2-ruby-bindings.spec

sbluhm:
rpmbuild fails without the fix. Corrected if statement to work with rpmbuild.


jreidinger (approved):
thanks for patch

PR #253, opened 2020-10-29

schubi2:
- https://trello.com/c/Vg5AS4wI
- https://bugzilla.suse.com/show_bug.cgi?id=1174198
Bildschirmfoto_2020-10-29_18-35-57

### DIFF
--- a/package/yast2-ruby-bindings.changes
+++ b/package/yast2-ruby-bindings.changes
@@ -3,7 +3,7 @@ Thu Oct 29 11:30:49 UTC 2020 - Stefan Schubert <schubi@suse.com>
 
 - Reporting augeas parsing errors and displaying them in rich-text
   format (bsc#1174198).
-- 4.2.8
+- 4.2.9
 
 -------------------------------------------------------------------
 Wed Jan 22 15:04:11 UTC 2020 - Ancor Gonzalez Sosa <ancor@suse.com>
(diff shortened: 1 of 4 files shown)


package/yast2-ruby-bindings.changes
```diff
@@ -1,3 +1,10 @@
+-------------------------------------------------------------------
+Thu Oct 29 11:30:49 UTC 2020 - Stefan Schubert <schubi@suse.com>
+
+- Reporting augeas parsing errors and displaying them in rich-text
+  format (bsc#1174198).
+- 4.2.8
```
mvidner, line 6:
```suggestion
- 4.2.9
```

src/ruby/yast/wfm.rb
```diff
@@ -4,6 +4,9 @@
 require "yast/debugger"
 require "yast/profiler"
 require "yast/yast"
+require "cgi"
+
+include CGI::Escape
```
mvidner, line 9: This would put `escapeHTML` into the global namespace. it is better to limit it to Yast::WFM.

`include` works for instance methods, `extend` for class methods

```rb
module Yast
  module WFM
    extend CGI::Escape
```

mvidner (changes requested):
Yay! Just final details:

src/ruby/yast/wfm.rb
jreidinger: I would not mess global namespace and neither WFM namespace and simply just call full `CGI.escape_html` where it is needed.

mvidner: @jreidinger thanks! I did not find that in the unofficial docs :( https://www.rubydoc.info/stdlib/cgi

 `escape_html` has nicer spelling but is not documented in the official docs: https://ruby-doc.org/stdlib-2.7.2/libdoc/cgi/rdoc/CGI.html

mvidner (approved):

PR #254, opened 2020-10-29
WIP: UI Unit Tests for MultiSelectionBox

shundhammer:
## Trello

https://trello.com/c/ZbkIbc0M/2133-3-probable-libyui-regressions-p1-1177985-and-p1-1177982

## New UI Unit Test

This tests some basic features of the MultiSelectionBox as well as the regression that was just fixed in https://github.com/libyui/libyui-ncurses/pull/108 .

It uses this new UI example from https://github.com/libyui/libyui-ncurses/pull/108 :

MultiSelectionBox-test-ncurses

https://github.com/yast/yast-ycp-ui-bindings/blob/master/examples/MultiSelectionBox-test.rb

# Problems

- This is not  100% stable. Sometimes some individual tests fail (typically the keyboard handling tests); it appears that this is because of screen update sync problems.

- Probably similar reason: After switching the `[ ] Vegetarian` check box, it should also select one more topping. That works when starting the example interactively, but never in the unit tests. This may also be a sync problem.

**_FIXME: TO DO_**

### DIFF
--- a/package/yast2-ruby-bindings.changes
+++ b/package/yast2-ruby-bindings.changes
@@ -1,4 +1,10 @@
 -------------------------------------------------------------------
+Thu Oct 29 17:26:31 UTC 2020 - Stefan Hundhammer <shundhammer@suse.com>
+
+- Added unit tests for NCMultiSelectionBox (bsc#1177985)
+- 4.3.6
+
+-------------------------------------------------------------------
 Tue Oct 13 14:42:52 UTC 2020 - Martin Vidner <mvidner@suse.com>
 
 - Add automatic TUI (ncurses) tests using tmux (bsc#1165388).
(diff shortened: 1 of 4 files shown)


shundhammer:
@mvidner, please have a look. Maybe you can spot something that I am doing wrong here.

mvidner:
In my TW VM, the test works every time so far, run as `rspec -f d tests/libyui/multi_selection_box_spec.rb`.

mvidner:
OK, in `rake osc:build` tests randomly fail and I think it is because we run these 2 commands without any delay:

```rb
        tui.send_key("Enter")
        expect(tui.capture_pane).to include("new content")
```

tmux captures the screen before the program has had time to react to the key press.

My solution is to add a 300 ms delay to the beginning of  `capture_pane`.

mvidner:
I have also checked that without the libyui-ncurses fix (https://github.com/libyui/libyui-ncurses/pull/108/) only the corresponding test fails as it should.

mvidner:
Continuing in #256

PR #255, opened 2020-10-30

schubi2:
A merge with https://github.com/yast/yast-ruby-bindings/pull/253


mvidner:
The travis failure is weird, an earlier ncurses test seems to hit internal libyui problems. But I am confident it is not this PR's fault. We have other PRs pending to investigate that.

mvidner (approved):

PR #256, opened 2020-10-30
UI Unit Tests for MultiSelectionBox

mvidner:
Continues #254.

@shundhammer if you make your PRs from a branch in the team repo instead of from your fork then we will be able to push to that branch and keep the same PR instead of needlessly creating a new one. thanks :smiley: 

Copying that PR's description, minus the *Problems* section since they are solved in this PR.

## Trello

https://trello.com/c/ZbkIbc0M/2133-3-probable-libyui-regressions-p1-1177985-and-p1-1177982

## New UI Unit Test

This tests some basic features of the MultiSelectionBox as well as the regression that was just fixed in https://github.com/libyui/libyui-ncurses/pull/108 .

It uses this new UI example from https://github.com/libyui/libyui-ncurses/pull/108 :

MultiSelectionBox-test-ncurses

https://github.com/yast/yast-ycp-ui-bindings/blob/master/examples/MultiSelectionBox-test.rb

### DIFF
--- a/package/yast2-ruby-bindings.changes
+++ b/package/yast2-ruby-bindings.changes
@@ -1,4 +1,10 @@
 -------------------------------------------------------------------
+Thu Oct 29 17:26:31 UTC 2020 - Stefan Hundhammer <shundhammer@suse.com>
+
+- Added unit tests for NCMultiSelectionBox (bsc#1177985)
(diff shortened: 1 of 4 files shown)


shundhammer (approved):
LGTM

PR #260, opened 2020-11-03
Test: menu items remain disabled after hotkeys are recomputed

mvidner:
- bug https://bugzilla.opensuse.org/show_bug.cgi?id=1178394
- the fix is https://github.com/libyui/libyui-qt/pull/139
- half of test case https://github.com/yast/yast-ycp-ui-bindings/pull/60
- this is the other half

Well, sort of. The bug is in libyui-qt and we test libyui-ncurses. Defense in depth.

Includes enabling the test for https://bugzilla.suse.com/show_bug.cgi?id=1177145 which got fixed together with https://bugzilla.suse.com/show_bug.cgi?id=1176402.

### DIFF
--- a/tests/libyui/menu_item_spec.rb
+++ b/tests/libyui/menu_item_spec.rb
@@ -27,4 +27,33 @@ describe "Menu Item" do
 
     @tui.send_keys "M-Q"        # &Quit
   end
+
+  bug = "1178394" # https://bugzilla.suse.com/show_bug.cgi?id=1178394
+  it "remains disabled after hotkeys are recomputed" do
+    @base = "menu_disabled_#{bug}"
+    @tui.await(/File.*Edit.*View/)
+    @tui.capture_pane_to("#{@base}-1-initial")
+
+    @tui.send_keys "M-E"        # &Edit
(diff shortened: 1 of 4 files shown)


tests/libyui/menu_item_spec.rb
```diff
@@ -0,0 +1,59 @@
+require_relative "rspec_tmux_tui"
+
+describe "Menu Item" do
+  around(:each) do |ex|
```
shundhammer, line 4: In this case it's only two tests, so it shouldn't matter that much. But in general: Is it a good idea to restart the program for every test? As tests accumulate, this will be a heavy performance hit.

In many cases, this one included, we can easily start the program once, execute many tests, and terminate it after the last one; provided, of course, that the individual tests are independent enough not to change the internal state of the test program too much.

Let's discuss this later today or maybe tomorrow.

shundhammer (approved):
LGTM

But we need to discuss our testing strategy; see my other comment.

tests/libyui/menu_item_spec.rb
mvidner: OK, good point. When we add new tests we should do it without program restarting, just to make sure the library makes it easy.

PR #267, opened 2021-02-01
Logging all available products information into directory /var/log/Ya…

schubi2:
…ST2/product_info after YaST has crashed.

### DIFF
--- a/src/ruby/yast/wfm.rb
+++ b/src/ruby/yast/wfm.rb
@@ -310,8 +310,10 @@ module Yast
       end
 
       # Logging all information about the product evaluation
-      Yast.import "ProductEvaluation"
-      ProductEvaluation.write("internal_error")
+      require "installation/installation_info"
+      ::Installation::InstallationInfo.instance.write(
+        "internal_error", { error: "Exception raised in #{client}",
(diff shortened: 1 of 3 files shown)


src/ruby/yast/wfm.rb
```diff
@@ -308,6 +308,11 @@ def self.call_builtin_wrapper(*args)
         Yast.import "Report"
         Report.LongError(msg, height: height)
       end
+
+      # Logging all information about the product evaluation
+      Yast.import "ProductEvaluation"
+      ProductEvaluation.write("internal_error")
```
lslezak, line 314: Maybe we should also write some message into the file describing the problem:
```suggestion
      ProductEvaluation.write("internal_error", { error: "Exception raised in #{client}",
        error_details: msg, error_backtrace: e.backtrace })
```

Just simply merge the optional extra keys passed to the function.

lslezak:
Actually I'm not sure if we should write the file in the global exception handler. The problem is that it might be triggered also in installed system, the `require`d file might be missing, etc...

It can potentially trigger some more problems. I'd close it, let see if we really need something like this...

PR #270, opened 2021-03-04
Change the special keybard shortcut

lslezak:
- Start a graphical menu instead of a low level command line console
- The command line console can be started from there as well
- Related to https://github.com/yast/yast-installation/pull/905
- Related to [jsc#PM-1895](https://jira.suse.com/browse/PM-1895), [jsc#SLE-16263](https://jira.suse.com/browse/SLE-16263)
- Tested manually in installed system
- 4.3.13

### DIFF
--- a/src/binary/Yast.cc
+++ b/src/binary/Yast.cc
@@ -306,7 +306,7 @@ static VALUE rescue_require_console(...)
 }
 
 /**
- * Start the configuration console, it calls "Installation::Console.run" Ruby code.
+ * Start the configuration console, it calls "Installation::Console::Menu.run" Ruby code.
  */
 static void start_config_console()
 {
--- a/src/binary/Yast.cc
+++ b/src/binary/Yast.cc
@@ -295,7 +295,7 @@ static void start_ruby_debugger()
 
 static VALUE require_console(...)
 {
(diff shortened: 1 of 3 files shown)


src/binary/Yast.cc
```diff
@@ -319,10 +319,11 @@ static void start_config_console()
     // LoadError caught
     if (success == Qfalse) return;
 
-    // call "Installation::Console.run"
+    // call "Installation::Console::Menu.run"
     VALUE installation = rb_const_get(rb_cObject, rb_intern("Installation"));
     VALUE console = rb_const_get(installation, rb_intern("Console"));
-    rb_funcall(console, rb_intern("run"), 0);
+    VALUE menu = rb_const_get(console, rb_intern("Menu"));
+    rb_funcall(menu, rb_intern("run"), 0);
```
jreidinger, line 326: please update documenation of method.

jreidinger: Also what happens when exception is raised?

lslezak: Updated.

When an exception is raised then it is caught by the global exception handler (that "internal error" popup), just like with other code.

jreidinger (approved):

PR #272, opened 2021-03-17
Fixed password encrypting functions in SLE15-SP2/Leap 15.2 and older

lslezak:
### Problem 

The encrypting builtins return `nil` in SLE15-SP2/Leap 15.2 and older, the unit tests fail with
```
Failures:

   1) Yast::Builtins.crypt works as expected
      Failure/Error: expect(res).to be_a String
        expected nil to be a kind of String
```

- See https://build.suse.de/package/live_build_log/Devel:YaST:Head/yast2-ruby-bindings/SUSE_SLE-15-SP2_GA/aarch64
- Or https://build.opensuse.org/package/live_build_log/YaST:Head/yast2-ruby-bindings/openSUSE_Leap_15.2/x86_64
- It turned out that in the older distros the crypt function returns `EINVAL` error when the passed entropy is `NULL`. In that case we need to explicitly pass some entropy.
- Partially reverts #250
- Related to bsc#1176924
- Tested in Leap 15.1

### DIFF
--- a/src/binary/Builtin.cc
+++ b/src/binary/Builtin.cc
@@ -225,7 +225,7 @@ extern "C" {
     char* retval = crypt_gensalt_ra (crypt_prefix, crypt_rounds, NULL, 0);
 
     // auto entropy might not be supported in some older systems (15.2 and older),
-    // if we get the "Invalid argument" error then read some entropy from
+    // if we get the EINVAL "Invalid argument" error then read some entropy from
     // /dev/urandom and try again
     if (!retval && errno == EINVAL)
     {
@@ -240,9 +240,11 @@ extern "C" {
 
       char entropy[16];
       const size_t entropy_len = sizeof(entropy);
-      if (read_loop (fd, entropy, entropy_len) != entropy_len)
+      int read_size = read_loop (fd, entropy, entropy_len);
+      close (fd);
+
+      if (read_size != entropy_len)
       {
-        close (fd);
         y2error ("Unable to obtain entropy from %s\n", device);
         return 0;
       }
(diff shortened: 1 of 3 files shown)


src/binary/Builtin.cc
```diff
@@ -222,16 +222,37 @@ extern "C" {
...
-    const char *entropy = NULL;
-    const size_t entropy_len = 0;
+    char* retval = crypt_gensalt_ra (crypt_prefix, crypt_rounds, NULL, 0);
 
-    char* retval = crypt_gensalt_ra (crypt_prefix, crypt_rounds, entropy, entropy_len);
-
-    if (!retval)
+    // auto entropy might not be supported in some older systems (15.2 and older),
+    // if we get the "Invalid argument" error then read some entropy from
+    // /dev/urandom and try again
```
jreidinger, line 229: maybe link this pr there? as there is no bugzilla, but it is good source of information

src/binary/Builtin.cc
```diff
@@ -222,16 +222,37 @@ extern "C" {
...
+      {
+        y2error ("Can't open %s for reading: %s\n", device, strerror (errno));
+        return 0;
+      }
+
+      char entropy[16];
+      const size_t entropy_len = sizeof(entropy);
+      if (read_loop (fd, entropy, entropy_len) != entropy_len)
+      {
+        close (fd);
```
jreidinger, line 245: why close fd only if read failed? does not it leak fd?

lslezak: Ah, good point... I just copied the original code...

src/binary/Builtin.cc
lslezak: I do not think it is needed, the comment above is enough IMHO. And the PR can be found from the GitHub history if needed...

jreidinger (approved):

PR #273, opened 2021-03-23
Add type checking with Sorbet

mvidner:
(Part of SUSE Hackweek: https://hackweek.suse.com/20/projects/type-check-yast-with-sorbet)

https://sorbet.org/ is a static (ahead of time) type checker.

### How it works in general 

Code is annotated with type declarations. Sorbet declarations are valid Ruby syntax and so can be inlined within the code or placed in separate files (RBI, for Ruby Interface).

The type checking is *gradual*, meaning that each file declares with a `# typed: level` header how strictly typed its code is: `ignore` is unchecked, `false` only checks constants (class names), `true` is the general level of checking (method names, argument counts, variable types) and there are `strict` and `strong` levels for even more checks.

The contents of the `sorbet/` directory is autogenerated.

### How it works in YaST

Manually written type declarations go to the `rbi/` directory and are installed in `/usr/share/YaST2/rbi` to be used by dependent packages.

### In yast2-ruby-bindings

This being a basic package used everywhere else, I suppose we will want very high level of type checking eventually.

As the first step, my goal was to add just enough types to enable detecting an (artificial) bug in yast2.rpm: Annotating `Stage.rb` (a very short file) just enough to detect that another file has a typo, `Stage.inital` instead of `Stage.initial`


mvidner:
I've put this code to https://build.opensuse.org/package/show/home:mvidner:branches:YaST:Head/yast2-ruby-bindings

rbi/yast/builtins.rbi
```diff
@@ -0,0 +1,61 @@
+# typed: strong
+
+module Yast::Builtins
```
jreidinger, line 3: Hi interesting, code few questions: Is there check that verify that rbi and real rb code matches?

rbi/yast/client.rbi
```diff
@@ -0,0 +1,9 @@
+# typed: strong
+
+module Yast
+  class Client
+    include I18n
```
jreidinger, line 5: is this needed? Cannot it load just from rb file?

rbi/yast/i18n.rbi
```diff
@@ -0,0 +1,15 @@
+# typed: strong
+
+# Namespacing note:
+# if we say `module Yast::I18n` then String means ::String
+# if we said `module Yast; module I18n` then String would mean Yast::String
```
jreidinger, line 5: maybe it would be better to be explicit and just use ::String?

rbi/yast/scr.rbi
```diff
@@ -0,0 +1,20 @@
...
+  end
+  def self.Execute(scr_path, arg1 = T.unsafe(nil), arg2 = T.unsafe(nil)); end
+
+  sig do
+    params(
+      scr_path: T.any(String, Yast::Path),
+      arg1: T.untyped
+    ).returns(T.untyped)
+  end
+  def self.Read(scr_path, arg1 = T.unsafe(nil)); end
```
jreidinger, line 19: Write is missing due to WIP or is there other reason?

rbi/yast/uishortcuts.rbi
```diff
@@ -0,0 +1,152 @@
+# typed: strong
+module Yast
+  module UIShortcuts
```
jreidinger, line 3: this looks quite fragile, maybe generate it by script? Also we basically kill type checking here as each type has own limitation for params

rbi/yast/wfm.rbi
```diff
@@ -0,0 +1,24 @@
+# typed: strong
+module Yast::WFM
+  class NoParameter; end
```
jreidinger, line 3: this looks quite magic for me. For sure it will need some explanation

rbi/yast/wfm.rbi
```diff
@@ -0,0 +1,24 @@
...
+  YCPAny = T.type_alias do
+    T.any(
+      NilClass,
+      T::Boolean,
+      Integer,
+      Float,
+      Symbol,
+      String,
+      T::Array[T.untyped],
```
jreidinger, line 12: cannot we have recursion here like `T::Array[YCPAny]`

tests/i18n_spec.rb
```diff
@@ -1,13 +1,20 @@
...
 
 require "yast"
 
-include Yast::I18n
 
-module Yast
-  describe I18n do
+
+describe Yast do
+  describe Yast::I18n do
```
jreidinger, line 11: this looks string, I prefer just `describe Yast::I18n" without that Yast one.

tests/i18n_spec.rb
```diff
@@ -1,13 +1,20 @@
...
+    include Yast::I18n
+    extend Yast::I18n
+
+    let(:translated) { "translated".freeze }
+    let(:singular) { "untranslated".freeze }
+    let(:plural) { "plural".freeze }
```
jreidinger, line 17: why constants does not work?

rbi/yast/builtins.rbi
mvidner: Good point!

- If I make a mistake in builtins.rbi, it will go undetected because builtins.rb is still `typed: false`
- convert.rb is `typed: true` but it defines its methods with metaprogramming so again, undetected :(
- ops.rb is `typed: false` but if I change it to `true` there are few enough errors and one of them is in fact a mismatch that you ask about:

```
src/ruby/yast/ops.rb:113: Method Yast::Ops.get redefined without matching argument count. Expected: 3, got: 4 https://srb.help/4010
     113 |    def self.get(object, indexes, default = nil, skip_frames = 0)
              ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
    rbi/yast/ops.rbi:15: Previous definition
    15 |  def self.get(object, indexes, default=T.unsafe(nil), &block); end
          ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
```

rbi/yast/client.rbi
mvidner: Sorbet takes great care to be fast so I guess the major reason is to be able to read just the RBI files of your dependencies and not parse and analyze all their sources.

rbi/yast/i18n.rbi
mvidner: Yast::String should be destroyed, it is a design bug in YCP Killer.
Do you really want to write `::String` instead of `String` everywhere?
Fortunately the two have such different APIs that an accidental confusion is resolved pretty quickly.

rbi/yast/scr.rbi
mvidner: Yup, just the minimal viable prototype. SCR.Write is pretty high in the TODO list.

rbi/yast/uishortcuts.rbi
mvidner: The way forward is to add more type information here, so generating this is a dead end.

So why did I include what appears to be a useless list?
It's because UIShortcuts is a mixin, and when I remove this RBI file, any `typed: true` class that includes it will fail on unknown methods. (seen in yast2.rpm)

rbi/yast/wfm.rbi
mvidner: Hm, you are right, I did perpetrate magic here. Fortunately it did not have time to mutate so I can still remember the reason and will document it.

```rb
#  See Args below
class NoParameter
...
# The following sig is a way to describe an optional parameter which has NOT a default value.
# It is a union of
#  sig { params().returns(T::Array[YCPAny]) }
#  sig { params(i: Integer).returns(YCPAny) }
```

rbi/yast/wfm.rbi
mvidner: Let's try... nope.

tests/i18n_spec.rb
mvidner: Sorbet is stricter than Ruby and thinks that constants defined in one `describe` are not visible in a *sibling* `describe`. I will commit a shorter fix which just moves the constants up one block level.

jreidinger: yeap, that will look better

PR #275, opened 2021-06-11
avoid crash caused by missing client argument

jreidinger:
bsc: https://bugzilla.suse.com/show_bug.cgi?id=1187230

Found during manual testing. Not sure about the best solution as result is still exception and internal error ( and it is really our internal fault, but maybe we can present it nicer? )

### DIFF
--- a/src/ruby/yast/wfm.rb
+++ b/src/ruby/yast/wfm.rb
@@ -326,14 +326,15 @@ module Yast
       log.info "To abort the process was requested from client #{client}: #{e.class}: #{e.message}"
     end
 
-    private_class_method def self.check_client_result_type!(result)
+    # @param client [String]
+    private_class_method def self.check_client_result_type!(result, client)
       allowed_types = Ops::TYPES_MAP.values.flatten
       allowed_types.delete(::Object) # remove generic type for any
 
       # check if response is allowed
       allowed = allowed_types.any? { |t| result.is_a? t }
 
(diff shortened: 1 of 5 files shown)


src/ruby/yast/wfm.rb
```diff
@@ -326,7 +326,8 @@ def self.call_builtin_wrapper(*args)
       log.info "To abort the process was requested from client #{client}: #{e.class}: #{e.message}"
     end
 
-    private_class_method def self.check_client_result_type!(result)
+    # @param client [String]
+    private_class_method def self.check_client_result_type!(result, client)
```
jreidinger, line 330: crash was at https://github.com/yast/yast-ruby-bindings/pull/275/files#diff-6890c75b6b9d4cea8eca284adb2f9a75a09810344ce90bf38a1d33394829692cR337

mvidner (changes requested):
Thanks!
It's funny how this is a familiar pattern of bugs: when something goes wrong and we are supposed to raise a meaningful exception, we instead crash.

1. Which brings me to a need for a unit test that checks the exception message.

2. Also, the messages can get better:

```rb
raise "Non-YCP type #{result.class} returned from client #{client}" unless allowed
```

commit message:
> When a WFM client returns a value not representable in YCP, raise a meaningful exception
> instead of crashing on an undefined variable

mvidner (approved):
Yay! Thank you!

PR #276, opened 2021-07-21
Avoid IO exception

joseivanlopez:
## Problem

When a signal is sent to YaST (e.g., when the parent terminal is killed), YaST tries writing a message to the standard error output. But stderr can be closed at that time, making YaST to raise an exception.

* https://bugzilla.suse.com/show_bug.cgi?id=1154854

## Solution

Avoid IO error by doing "exception swallowing"  for *Errno::EIO* while writing to stderr.

See also https://github.com/yast/yast-control-center/pull/51.

## Testing

Manually tested.

### DIFF
--- a/src/ruby/yast/y2start_helpers.rb
+++ b/src/ruby/yast/y2start_helpers.rb
@@ -137,7 +137,13 @@ module Yast
     private_class_method def self.signal_handler(name)
       Signal.trap(name, "IGNORE")
 
-      $stderr.puts "YaST got signal #{name}."
+      # Exception swallowing: writing to stderr could fail if the parent process was killed,
+      # see bsc#1154854.
+      begin
+        $stderr.puts "YaST got signal #{name}."
+      rescue Errno::EIO
+        # Nothing to do
+      end
 
       signal_log_open do |f|
         f.puts "=== #{Time.now} ==="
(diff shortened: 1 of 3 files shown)


src/ruby/yast/y2start_helpers.rb
```diff
@@ -137,7 +137,13 @@ def self.generate_exit_code(value)
     private_class_method def self.signal_handler(name)
       Signal.trap(name, "IGNORE")
 
-      $stderr.puts "YaST got signal #{name}."
+      # Exception swallowing: writing to stderr could fail if the parent process was killed,
+      # see bsc#1154854.
+      begin
+        $stderr.puts "YaST got signal #{name}."
```
jreidinger, line 143: is in that case io already closed? If so, we can even do it without exception with `$stderr.puts "YaST got signal #{name}." unless $stderr.closed?`

joseivanlopez: Nop, is not closed :(, at least the method returns *false*.

jreidinger (approved):

src/ruby/yast/y2start_helpers.rb
jreidinger: so maybe mention it in comment? I expect I am not alone who ask

PR #277, opened 2021-08-13
Added empty help test (bsc#972548)

lslezak:
## Improved Tests

- The `tests/integration/run.rb` script used fixed temporary file names in `/tmp`, that's a security problem
  - Switched to the `Tempfile` objects
  - The temporary files are automatically created, so no need to check the existence
- Updated the libyui tests to dump the screenshots to the `log/` subdirectory (to not mess up the sources)
- Added unit test for https://bugzilla.suse.com/show_bug.cgi?id=972548

PR #278, opened 2021-08-18
Fix deps

jreidinger:
related to failure on s390 from https://github.com/yast/yast-ruby-bindings/pull/277

### DIFF
--- a/package/yast2-ruby-bindings.spec
+++ b/package/yast2-ruby-bindings.spec
@@ -17,7 +17,7 @@
 
 
 Name:           yast2-ruby-bindings
-Version:        4.4.3
+Version:        4.4.4
 Release:        0
 URL:            https://github.com/yast/yast-ruby-bindings
 BuildRoot:      %{_tmppath}/%{name}-%{version}-build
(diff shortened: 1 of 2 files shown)


package/yast2-ruby-bindings.spec
```diff
@@ -28,21 +28,23 @@ BuildRequires:  cmake
...
 BuildRequires:  ruby-devel
 Requires:       yast2-core >= 3.2.2
 BuildRequires:  yast2-core-devel >= 3.2.2
 # MenuBar-shortcuts-test.rb
 Requires:       yast2-ycp-ui-bindings       >= 4.3.7
 BuildRequires:  yast2-ycp-ui-bindings-devel >= 4.3.7
+%ifarch s390 s390x
+# s390 specific frame title that is read from readvalues from s390-tools
+# needed also for tests, so build require it
+BuildRequires:  s390-tools
```
lslezak, line 43: Alternatively we could fix the [read_values](https://github.com/yast/yast-ruby-bindings/blob/a66796909dfa0ab42642e9969244622f08ca8b02/src/ruby/yast/y2start_helpers.rb#L125) method to handle missing `read_values` binary.

The advantage would be smaller build dependencies, during test we do not care much about the machine title...

But I guess the `s390-tools` package is small and it is an acceptable dependency.

lslezak (approved):

package/yast2-ruby-bindings.spec
jreidinger: yeap and also quite mandatory on s390. It quite surprise me that it is not already installed in build env.
And I prefer to not touch production code for test cases. Only if we want to handle missing s390-tools also outside of test, but there I prefer consistency with having title always.

PR #279, opened 2021-11-29
Ruby3

jreidinger:
Issue: ruby 3.0 produce warning:
```
/home/abuild/rpmbuild/BUILD/yast2-ruby-bindings-4.4.4/src/ruby/yast/ui.rb:4: warning: finalizer references object to be finalized
```

Fix:
store reference early, so it does not warn when it is finalized


lslezak (approved):
LGTM

PR #281, opened 2022-01-17
Added yast/rspec/helpers.rb (related to bsc#1194784)

lslezak:
- See https://github.com/yast/yast-packager/pull/599#pullrequestreview-853381879
- Let's share the helper method
- It can also replace the existing stubs like
```ruby
def stub_module(name)
  Yast.const_set(name.to_sym, Class.new { def self.fake_method; end })
end

stub_module("Proxy")
```
with 
```ruby
Yast::RSpec::Helpers.define_yast_module("Proxy")
```

### DIFF
--- a/src/ruby/yast/rspec.rb
+++ b/src/ruby/yast/rspec.rb
@@ -1,6 +1,7 @@
+require "yast/rspec/helpers"
+require "yast/rspec/matchers"
 require "yast/rspec/scr"
 require "yast/rspec/shortcuts"
-require "yast/rspec/matchers"
 
 RSpec.configure do |c|
   c.include Yast::RSpec::Shortcuts
--- a/src/ruby/yast/rspec/helpers.rb
+++ b/src/ruby/yast/rspec/helpers.rb
@@ -43,7 +43,9 @@ module Yast
       #   verifying doubles!
       # @param block [Block] optional method definitions, they should provide
       #   the same API as the original module, this can be combined with the
-      #   "methods" parameter
+      #   `methods` parameter. The block is evaluated in the context of the
+      #   defined modules so you can also use the helpers like `attr_reader`
+      #   or define constants.
       def self.define_yast_module(name, methods: [:fake_method], force: false, &block)
         # sanity check, make sure the code coverage is already running if it is enabled
         if ENV["COVERAGE"] && (!defined?(SimpleCov) || !SimpleCov.running)
--- a/src/ruby/yast/rspec/helpers.rb
+++ b/src/ruby/yast/rspec/helpers.rb
@@ -10,8 +10,11 @@ module Yast
       # as usually. If it is missing a substitute definition is created.
       #
       # @note This needs to be called *after* starting code coverage with
-      # `SimpleCov.start` so the coverage of the imported modules is correctly
-      # counted.
+      #   `SimpleCov.start` so the coverage of the imported modules is correctly
+      #   counted.
+      #
+      # @note You can force using the defined stubs although the modules are
+      #   present in the system, this might be useful to actually test the stubs.
       #
       # @example Mock the `Language` module and the `Language.language` method
       # Yast::RSpec::Helpers.define_yast_module("Language") do
@@ -40,11 +43,19 @@ module Yast
             "  Called from: #{caller(1).first}\n\n"
         end
 
-        # try loading the module, it might be present in the system (running locally
-        # or in GitHub Actions), mock it only when missing (e.g. in OBS build)
-        Yast.import(name)
-        puts "Found module Yast::#{name}"
+        if ENV["YAST_FORCE_MODULE_STUBS"]
+          define_missing_yast_module(name, &block)
+        else
+          # try loading the module, it might be present in the system (running locally
+          # or in GitHub Actions), mock it only when missing (e.g. in OBS build)
+          Yast.import(name)
+          puts "Found module Yast::#{name}"
+        end
       rescue NameError
+        define_missing_yast_module(name, &block)
+      end
+
+      private_class_method def self.define_missing_yast_module(name, &block)
         warn "Module Yast::#{name} not found"
         if block_given?
           block.call
--- a/src/ruby/yast/rspec/helpers.rb
+++ b/src/ruby/yast/rspec/helpers.rb
@@ -20,21 +20,22 @@ module Yast
       #   present in the system, this might be useful to actually test the stubs.
       #   Just define the YAST_FORCE_MODULE_STUBS environment variable.
       #
-      # @example Mock empty `AutoInstall` module
-      # Yast::RSpec::Helpers.define_yast_module("AutoInstall")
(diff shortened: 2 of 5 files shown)


src/ruby/yast/rspec/helpers.rb
```diff
@@ -0,0 +1,69 @@
...
+          # or in GitHub Actions), mock it only when missing (e.g. in OBS build)
+          Yast.import(name)
+          puts "Found module Yast::#{name}"
+        end
+      rescue NameError
+        define_missing_yast_module(name, &block)
+      end
+
+      private_class_method def self.define_missing_yast_module(name, &block)
+        warn "Module Yast::#{name} not found"
```
dgdavid, line 59: NP: I'd say "Mocking not found Yast::#{name} module"

dgdavid (approved):
LGTM :+1:

src/ruby/yast/rspec/helpers.rb
```diff
@@ -0,0 +1,69 @@
...
+      #   methods used in the tests, if no block is passed an empty module is defined
+      def self.define_yast_module(name, &block)
+        # sanity check, make sure the code coverage is already running if it is enabled
+        if ENV["COVERAGE"] && (!defined?(SimpleCov) || !SimpleCov.running)
+          abort "\nERROR: The `define_yast_module` method needs to be called *after* " \
+            "enabling the code coverage tracking with `SimpleCov.start`!\n" \
+            "  Called from: #{caller(1).first}\n\n"
+        end
+
+        if ENV["YAST_FORCE_MODULE_STUBS"]
```
jreidinger, line 46: this should be documented with its use case.

src/ruby/yast/rspec/helpers.rb
```diff
@@ -0,0 +1,69 @@
...
+          define_missing_yast_module(name, &block)
+        else
+          # try loading the module, it might be present in the system (running locally
+          # or in GitHub Actions), mock it only when missing (e.g. in OBS build)
+          Yast.import(name)
+          puts "Found module Yast::#{name}"
```
jreidinger, line 52: puts is not nice in rspec output. If you need to print it, then I would do it only for verbose output.

src/ruby/yast/rspec/helpers.rb
```diff
@@ -0,0 +1,69 @@
...
+        define_missing_yast_module(name, &block)
+      end
+
+      private_class_method def self.define_missing_yast_module(name, &block)
+        warn "Module Yast::#{name} not found"
+        if block_given?
+          block.call
+        else
+          # create a fake implementation of the module
+          Yast.const_set(name.to_sym, Class.new { def self.fake_method; end })
```
jreidinger, line 64: when I see this, what also about having as optional argument list of methods to fake? so it can be something like `self.define_missing_yast_module(name, methods: [:fake_method], &block)`

Also block variant looks kind of strange for me. I would expect that return value of block can be assigned to that name or even block invoked in context of that new Class. which would be the best from my POV.

So code like
```
Object.const_set(name.to_sym, Class.new do
  methods.each do { |m| define_singleton_method(m) {} }
 instance_eval(&block)
end)
```

This way. you can have for you Language example
```
Yast::RSpec::Helpers.define_yast_module("Language", methods: [:language])
```
or if you need specific execution in method
```
Yast::RSpec::Helpers.define_yast_module("Language") do
  def language
     "en_US"
  end
end
```

and with include I mention above, you can omit whole `Yast::RSpec::Helpers` part.

src/ruby/yast/rspec.rb
```diff
@@ -1,6 +1,7 @@
+require "yast/rspec/helpers"
+require "yast/rspec/matchers"
 require "yast/rspec/scr"
 require "yast/rspec/shortcuts"
-require "yast/rspec/matchers"
 
 RSpec.configure do |c|
   c.include Yast::RSpec::Shortcuts
```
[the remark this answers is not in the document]
lslezak: The [Rspec include](https://rubydoc.info/github/rspec/rspec-core/RSpec/Core/Configuration#include-instance_method) adds the methods into the example groups.

For mocking YaST modules that is usually too late, you need to mock them *before* loading any test as the `Yast.import` is called very early. IIRC we mock all YaST modules in the `test_helper.rb` so the include would not make sense, it would be unused.

src/ruby/yast/rspec/helpers.rb
```diff
@@ -9,60 +9,76 @@ module Helpers
...
-      #   # see modules/Language.rb in yast2-country
-      #   module Yast
-      #     class LanguageClass < Module
-      #       # @return [String]
-      #       def language; end
-      #     end
-      #     Language = LanguageClass.new
-      #   end
-      # end
+      #   Just the define the YAST_FORCE_MODULE_STUBS environment variable.
```
jreidinger, line 21: typo "the define"

src/ruby/yast/rspec/helpers.rb
```diff
@@ -9,60 +9,76 @@ module Helpers
...
+      #   def Import(_config); end
+      # end
+      #
       # @param name [String] name of the YaST module
-      # @param block [Block] optional module definition, it should provide
-      #   the same API as the original module, it is enough to define only the
-      #   methods used in the tests, if no block is passed an empty module is defined
-      def self.define_yast_module(name, &block)
+      # @param methods [Array<Symbol>] optional list of defined methods,
+      #   the defined methods accept no parameter, if you need a parameter
```
jreidinger, line 36: maybe this can be easily fixed by 
```
          methods.each { |m| define_singleton_method(m) {|*_a, **_k, &_b| } }
```
this way it accepts everything

src/ruby/yast/rspec/helpers.rb
```diff
@@ -9,60 +9,76 @@ module Helpers
...
+      #   then define it in the block
+      # @param force [Boolean] force creating the fake implementation even when
+      #   the module is present in the system, can be also set by the
+      #   YAST_FORCE_MODULE_STUBS environment variable. This might be useful
+      #   if the module constructor touches the system and it would need a lot
+      #   of mocking. But use it carefully, this defeats the purpose of the RSpec
+      #   verifying doubles!
+      # @param block [Block] optional method definitions, they should provide
+      #   the same API as the original module
```
jreidinger, line 45: I would add note that it can  be combined with methods parameter. So simple methods can be created by param and tricky ones with block. And also that block is basically invoked in context of that class, so it allows attr_readers, constants and other stuff.

src/ruby/yast/rspec/helpers.rb
lslezak: IIRC the RSpec verifying doubles not only check that the mocked method exists but also checks the arity of the method. So that would very likely break with this change.

jreidinger (approved):

jreidinger:
Thanks for great patch. I think it is worth to announce it on yast-devel.

PR #283, opened 2022-02-07
Set "x509ignoreCN=0" GODEBUG environment variable (bsc#1195220)

lslezak:
- https://bugzilla.suse.com/show_bug.cgi?id=1195220
- Enable CN (Common Name) matching in SSL certificates in Go programs (e.g. `suseconnect-ng`). Without it the CN would be ignored and only the SAN (Subject Alternative Name) certificate fields would be used.
- But the self-signed certificates usually only contain the CN field without SAN so YaST could not connect to SMT/RMT
registration servers.
- This error prevented from importing the SSL certificate (that's enabled only for specific known errors): cert_cn_problem

## Test

- Tested manually, the certificate can be now imported: cert_cn_fixed

## Notes

- Originally I wanted to fix that in `yast2-registration` but it turned out that Go reads the setting only at startup, changing the environment variable later does not have any effect.
- During installation the registration code might be loaded early in the self-update code, we need to be sure it is set before loading `suseconnect-ng` from anywhere.
- It needs to be set also in installed system so it cannot be fixed in `yast2-installation` starting scrips.
- See also related fix https://github.com/SUSE/connect-ng/pull/118

### DIFF
--- a/src/ruby/yast/y2start_helpers.rb
+++ b/src/ruby/yast/y2start_helpers.rb
@@ -2,15 +2,30 @@ module Yast
   module Y2StartHelpers
     # Configure global environment for YaST
     #
-    # Currently it only sets values for $PATH.
+    # Currently it only sets values for $PATH and $GODEBUG.
     #
     # By configuring $PATH, it ensures that correct external programs are executed when
     # relative paths are given, so possible CVEs are avoided when running YaST.
+    #
+    # $GODEBUG is configured to enable CN (Common Name) matching in SSL certificates
+    # used by Go programs (suseconnect-ng used by registration)
     #
     # Note that forked processes will inherit the environment configuration, for example
     # when executing commands via SCR or Cheetah.
     def self.config_env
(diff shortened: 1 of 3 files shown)


src/ruby/yast/y2start_helpers.rb
```diff
@@ -2,15 +2,30 @@ module Yast
...
+    # $GODEBUG is configured to enable CN (Common Name) matching in SSL certificates
+    # used by Go programs (suseconnect-ng used by registration)
+    #
     # Note that forked processes will inherit the environment configuration, for example
     # when executing commands via SCR or Cheetah.
     def self.config_env
       ENV["PATH"] = "/sbin:/usr/sbin:/usr/bin:/bin"
+
+      # Note: this setting was removed in go-1.17 (https://go.dev/doc/go1.17),
+      # SLE15 uses go-1.16
```
jreidinger, line 19: I am missing in code also reference to bugzilla. Other than that it looks good.

jreidinger (approved):

PR #287, opened 2022-10-24
Added "log.group" method, allow installation via "yupdate"

lslezak:
## Problem

- https://bugzilla.suse.com/show_bug.cgi?id=1204625
- In short: YaST should allow log grouping similar to [GitHub log grouping](https://docs.github.com/en/actions/using-workflows/workflow-commands-for-github-actions#grouping-log-lines)

## Solution

Added new `log.group` method, see the [example log from TW installation](https://lslezak.github.io/ylogviewer/?log_url=https%3A%2F%2Fgist.githubusercontent.com%2Flslezak%2Fd36a2a15b9ccd49f035c7e51b4818ee5%2Fraw%2Fa8f2822f608f7ae0bbabb3dbe457b5202e21da25%2Fy2log-tw-installation.tar.xz) how it will work (click the "Load URL" button to load the example log).

```ruby
log.group "Description of the group" do
  foo
end
```

This logs special begin and end marks to the y2log, these might be used when reading/displaying the log.

It returns the block return value, it can be easily used like

```ruby
# returns the `bar` value
def foo
  log.group "Description of the group" do
    bar
  end
end
```

Optionally you can describe also the result

```ruby
  log.group "Description" do |group|
    ret = bar
    group.summary "Installed {ret} packages"
    ret
  end
```

The code automatically treats values `:abort`, `:cancel` and `false` as failures. But you can explicitly mark a failure

```ruby
  log.group "Description" do |group|
    ret = bar
    group.failed = true if ret == 0
    ret
  end
```
### `yupdate` Support

The `Rakefile` was updated so the Ruby files can be easily updated in the inst-sys with the `yupdate` command.

## Testing

- Added a new unit test
- Tested manually, see an [example log from TW installation](https://lslezak.github.io/ylogviewer/?log_url=https%3A%2F%2Fgist.githubusercontent.com%2Flslezak%2Fd36a2a15b9ccd49f035c7e51b4818ee5%2Fraw%2Fa8f2822f608f7ae0bbabb3dbe457b5202e21da25%2Fy2log-tw-installation.tar.xz) (click the "Load URL" button to load the example log)

### DIFF
--- a/src/ruby/yast/y2logger.rb
+++ b/src/ruby/yast/y2logger.rb
@@ -6,6 +6,41 @@ require "singleton"
 require "yast/logger"
 
 module Yast
+
+  # class for storing the log group result data, used in the Y2Logger.group
+  class LogGroupResult
+    # @!attribute [rw] result summary, human readable description
+    # @return [String,nil] result of the step as a textual description
+    attr_accessor :summary
+
+    # @!attribute [rw] result of the block
+    # @return result of the step
+    attr_accessor :result
+
+    # @!attribute [w] set to true if the result of the group is a failure,
+    #  overrides the state evaluated from the the `result` attribute
+    attr_writer :failed
+
+    # these return values are considered failures by default,
+    # can be overridden by setting the `failed` attribute
+    FAILURES = [:abort, :cancel, false]
+
+    # did the execution of the block failed?
+    # @return [Boolean] true if the block failed
+    def failed?
+      if failed.nil?
+        FAILURES.include?(result)
+      else
+        failed
+      end
+    end
+
+  private
+
+    # @return [Boolean,nil] explicit failure
+    attr_reader :failed
+  end
+
   # A Ruby Logger which wraps Yast.y2*() calls
   class Y2Logger < ::Logger
     include Singleton
@@ -40,6 +75,37 @@ module Yast
       # process also debug messages but might not be logged in the end
       self.level = ::Logger::DEBUG
     end
+
+    # log a block of commands, adds a special begin and end markers into the log,
+    # the block should be one big logical step in the process,
+    # can be used recursively, e.g. log.group might call another log.group inside
+    # @param description [String] short description of the block
+    # @param block [Proc] block to call
+    # @yield param group [LogGroupResult] can be optionally used to pass result details
+    def group(description, &block)
+      details = LogGroupResult.new
+      # mark start of the group
+      info "::group::#{Process.clock_gettime(Process::CLOCK_MONOTONIC)}::#{description}"
+
+      if block_given?
+        ret = block.call(details)
+      else
+        raise ArgumentError, "Missing block parameter"
+      end
+      details.result = ret
+
+      ret
+    rescue StandardError => e
+      # mark a failure
+      details.failed = true
+      details.summary = "Raised exception: #{e}"
+      # reraise the original exception
+      raise
+    ensure
+      # mark end of the group with result data, if it failed log as an error
+      level = details.failed? ? :error : :info
+      send(level, "::endgroup::#{Process.clock_gettime(Process::CLOCK_MONOTONIC)}::#{details.summary}")
+    end
   end
 
   # This module provides access to Yast specific logging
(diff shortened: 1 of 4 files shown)


src/ruby/yast/y2logger.rb
```diff
@@ -6,6 +6,41 @@
...
+    # @return result of the step
+    attr_accessor :result
+
+    # @!attribute [w] set to true if the result of the group is a failure,
+    #  overrides the state evaluated from the the `result` attribute
+    attr_writer :failed
+
+    # these return values are considered failures by default,
+    # can be overridden by setting the `failed` attribute
+    FAILURES = [:abort, :cancel, false]
```
jreidinger, line 26: I would make this configurable. So you can specify what is failure

src/ruby/yast/y2logger.rb
```diff
@@ -40,6 +75,37 @@ def initialize(*_args)
...
+    rescue StandardError => e
+      # mark a failure
+      details.failed = true
+      details.summary = "Raised exception: #{e}"
+      # reraise the original exception
+      raise
+    ensure
+      # mark end of the group with result data, if it failed log as an error
+      level = details.failed? ? :error : :info
+      send(level, "::endgroup::#{Process.clock_gettime(Process::CLOCK_MONOTONIC)}::#{details.summary}")
```
jreidinger, line 107: public_send?

src/ruby/yast/y2logger.rb
```diff
@@ -40,6 +75,37 @@ def initialize(*_args)
...
+    # @yield param group [LogGroupResult] can be optionally used to pass result details
+    def group(description, &block)
+      details = LogGroupResult.new
+      # mark start of the group
+      info "::group::#{Process.clock_gettime(Process::CLOCK_MONOTONIC)}::#{description}"
+
+      if block_given?
+        ret = block.call(details)
+      else
+        raise ArgumentError, "Missing block parameter"
```
mvidner, line 95: This way we raise ArgumentError *after* logging the start of the group, which is IMHO better for giving a context. :+1:

mvidner, line 95: "block parameter" is ambiguous, without context I could interpret it as a parameter **of** a block

```suggestion
        raise ArgumentError, "Missing a block"
```

src/ruby/yast/y2logger.rb
jreidinger: good. Explicit one is ok...BTW I am missing test for explicit failed and I also failed to find it in code. I only see `group.success =`

mvidner: That's because Lada has changed `failed` to `success` in the meantime, on my suggestion :)

src/ruby/yast/y2logger.rb
```diff
@@ -60,6 +133,23 @@ def initialize(*_args)
...
+  #          log.group("Adding repositories") do
+  #            add_repositories
+  #          end
+  #
+  #          # log a logical group of commands with result details
+  #          log.group("Adding repositories") do |group|
+  #            ret, repos = add_repositories
+  #            if ret == :failed
+  #              group.summary = "Could not add repositories"
+  #              group.failed = true
```
jreidinger, line 148: so this comment is no longer valid

jreidinger (approved):
doc NP

src/ruby/yast/y2logger.rb
mvidner:
```suggestion
  #              group.success = false
```

src/ruby/yast/y2logger.rb
```diff
@@ -40,6 +78,41 @@ def initialize(*_args)
...
     end
+
+    # log a block of commands, adds a special begin and end markers into the log,
+    # the block should be one big logical step in the process,
+    # can be used recursively, e.g. log.group might call another log.group inside
+    # @param description [String] short description of the block
+    # @param block [Proc] block to call
+    # @yieldparam group [LogGroupResult] can be optionally used to pass result details
+    # @yieldreturn [Object] passed on;
+    #   if one of `false`, `:abort`, `:cancel` ({LogGroupResult::FAILURES}),
```
mvidner, line 89:
```suggestion
    #   if one of `false`, `:abort`, `:cancel` ({LogGroupResult#error_values}),
```

PR #289, opened 2023-03-30
Optimize memory used by the `publish` calls

jreidinger:
## Problem

Agame and also yast consume quite bit of memory, so lets try to optimize it.

## Solution

Add some profiling code and based on it, apply some optimization.

Optimization results ( testing on Leap15.4 so with ruby 2.5 ) just totals:

Original result:

```
Total allocated: 33846260 bytes (365250 objects)
Total retained:  5387014 bytes (47980 objects)
```

after first optimization:

```
Total allocated: 32843452 bytes (364422 objects)
Total retained:  5383110 bytes (47979 objects)
```

after second optimization:

```
Total allocated: 31132062 bytes (351286 objects)
Total retained:  4337427 bytes (41452 objects)
```

So result is roughly saved **2650 KiB** of allocated and in persistent it is **1021 KiB** on testing  sample that just require yast and imports some yast modules.

### DIFF
--- a/package/yast2-ruby-bindings.changes
+++ b/package/yast2-ruby-bindings.changes
@@ -1,4 +1,10 @@
 -------------------------------------------------------------------
+Mon Apr  3 10:46:37 UTC 2023 - Josef Reidinger <jreidinger@suse.com>
+
+- Improve YaST memory consumption (bsc#1210051)
+- 4.6.2
+
+-------------------------------------------------------------------
 Thu Mar  9 10:36:23 UTC 2023 - Martin Vidner <mvidner@suse.com>
 
 - Use ruby-devel versioned to match the gems (bsc#1209098)
--- a/package/yast2-ruby-bindings.changes
+++ b/package/yast2-ruby-bindings.changes
@@ -1,7 +1,7 @@
 -------------------------------------------------------------------
 Mon Apr  3 10:46:37 UTC 2023 - Josef Reidinger <jreidinger@suse.com>
 
-- Improve YaST memory consumption (bsc#1210051)
+- Improve YaST memory consumption related to import+publish (bsc#1210051)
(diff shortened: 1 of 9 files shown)


mvidner:
BTW I've tried enabling frozen string literals to see what memory effect it has, but fast_gettext does not like it, starting here
```console
$ ruby --enable=frozen-string-literal -r ../tests/test_helper.rb ../profiling/yast_import.rb  |& o
/usr/lib64/ruby/gems/2.5.0/gems/fast_gettext-1.6.0/lib/fast_gettext/vendor/mofile.rb:52:in `force_encoding': can't modify frozen String (FrozenError)
        from /usr/lib64/ruby/gems/2.5.0/gems/fast_gettext-1.6.0/lib/fast_gettext/vendor/mofile.rb:52:in `<class:MOFile>'
...
```

https://github.com/grosser/fast_gettext/blob/a4628d516cfd511bef3acc0762d453934f5a9acd/lib/fast_gettext/vendor/mofile.rb#L52

mvidner (commented):
more improvements

mvidner:
@jreidinger later, I think it is worth looking at the **retained *foo*** part of the `memory_profiler` report, but for a real workload, not just for an initial import like your test

package/yast2-ruby-bindings.changes
```diff
@@ -1,3 +1,9 @@
+-------------------------------------------------------------------
+Mon Apr  3 10:46:37 UTC 2023 - Josef Reidinger <jreidinger@suse.com>
+
+- Improve YaST memory consumption (bsc#1210051)
```
mvidner, line 4:
```suggestion
- Improve YaST memory consumption related to import+publish (bsc#1210051)
```

mvidner (approved):

PR #291, opened 2023-09-08
Backport sp6

jreidinger:
## Problem

SLE 15 SP6 branch is based on SLE15 SP5 maintenance branch. So fixes done in master can be sometimes useful also for SP6.

## Solution

Backport both fixes (#288 and #289) , so it is same as 4.6.2.


mvidner (approved):
Thanks

But note that we had to fix some fallout which will have to be backported too: https://github.com/yast/yast-packager/pull/638

jreidinger:
@mvidner it is there at https://github.com/yast/yast-packager/pull/641

PR #295, opened 2025-05-13
Drop YaST Command Line Interface Support

shundhammer:
## Bugzilla

https://bugzilla.suse.com/show_bug.cgi?id=1243018

## Problem / Motivation

YaST is only used as a library for the Agama backend in SLES-16 / Leap 16.

The functionality of YaST goes much further, but  we cannot guarantee that all the details and all the behavior in YaST are compatible in every detail with SLES-16 / Leap 16. This will create pitfalls for users who may find a way to still use YaST functionality that will then increasingly be outdated.

In long run, this will be true for Factory / Tumbleweed as well.

So we are starting to prevent pitfalls with this: The YaST command line  interface (CLI) should not be accessible anymore. This is even more important because this is an aspect of YaST that is less commonly used and thus less thoroughly tested.

## Fix

This PR makes the YaST  CLI unaccessible: When trying to use it, it will simply exit with an error message in the log and on stderr.

```console
sh@balrog:~> sudo yast2 lan list
FATAL: The YaST command line interface is not supported anymore. Exiting.
sh@balrog:~>
```

```console
sh@balrog:~> sudo yast2 lan
```
(-> The YaST lan module starts normally)

Some few YaST modules accept one command line argument for GUI operation, i.e. outside of the YaST CLI mode:

```console
sh@balrog:~> sudo yast2 view_anymsg /etc/os-release
```
(-> The YaST view_anymsg module opens a window showing the `/etc/os-release` file)

### DIFF
--- a/package/yast2-ruby-bindings.changes
+++ b/package/yast2-ruby-bindings.changes
@@ -1,4 +1,10 @@
 -------------------------------------------------------------------
+Tue May 13 11:42:11 UTC 2025 - Stefan Hundhammer <shundhammer@suse.com>
+
+- Drop the YaST command line interface (CLI) (bsc#1243018)
+- 5.0.3
+
+-------------------------------------------------------------------
 Wed Jan 15 08:39:02 UTC 2025 - Josef Reidinger <jreidinger@suse.com>
 
 - Disable integration tests with libyui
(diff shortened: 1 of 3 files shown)


shundhammer:
This supersedes https://github.com/yast/yast-yast2/pull/1318.

jreidinger (approved):

PR #297, opened 2025-07-16
Allow building using CMake 4.0

DimStar77:
CMake 4.0 allows a minimum value of 3.5 for 'minimum_required_version'
Our CMakeLists.txt is simple enough to not need more changes than that.

CMake 3.5 was released on 2016, making it old enough to accept its general presence

## Problem

*Short description of the original problem.*

- *Bugzilla link*
- *openQA link*
- *Links to other related pull requests*

## Solution

*Short description of the fix.*

## Testing

- *Added a new unit test*
- *Tested manually*

## Screenshots

*If the fix affects the UI attach some screenshots here.*


joseivanlopez (approved):
LGTM

jreidinger:
Thanks for patch. Lets not block merge with our broken CI which we should fix ourself. ( missing fast gettext in our docket testing container )

